PR #18, opened 2023-09-15
Pan svg point

IongIer:


app/src/common/game_state.rs
IongIer, line 1: This was clippy

app/src/organisms/reserve.rs
IongIer, line 1: Clippy again

PR #70, opened 2023-11-16
Users games

IongIer:


Cargo.toml
```diff
@@ -20,7 +20,8 @@ leptos_icons = { version = "0.1", default_features = false, features = [
     "AiStepBackwardFilled",
     "AiFastForwardFilled",
     "AiStepForwardFilled",
-    "BiInfiniteRegular"
+    "BiInfiniteRegular",
+    "RiSwordOthersLine"
```
klautcomputing, line 24: Is there a better way to do this instead of pulling every single Icon in one by one?

apis/src/components/layouts/base_layout.rs
```diff
@@ -9,16 +9,15 @@ pub fn BaseLayout(children: Children) -> impl IntoView {
...
-            };
-            format!("{} {}", classes, theme)
+            }
         }/>
-        <Body class="h-full w-full bg-white dark:bg-gray-900"/>
+
+        <Body/>
         <Stylesheet id="leptos" href="/pkg/HiveGame.css"/>
-        <main class="h-full w-full">
+        <main class="min-h-screen h-full w-full bg-white dark:bg-gray-900">
```
klautcomputing, line 20: I won't pretend I understand what's going on here :D

apis/src/components/organisms/display_profile.rs
```diff
@@ -0,0 +1,97 @@
...
+use leptos::{logging::log, *};
+
+#[derive(Clone, PartialEq)]
+pub enum TabView {
+    Playing,
+    Games,
+}
+
+#[component]
+pub fn DisplayProfile(user: StoredValue<UserResponse>) -> impl IntoView {
```
klautcomputing, line 15: I would swap the location of playing and finished so that playing is left and finished games are on the right.

klautcomputing (approved):
Nitpicks, fix or ignore :+1:

PR #111, opened 2024-01-05
Add flake setup

LucianU:
I used docker instead of buildah, because the latter wouldn't build (the irony) on my M1 Mac.

Let me know if anything is unclear.


LucianU:
Closing, as a more detailed flake was implemented in #168.

PR #119, opened 2024-01-10
Navigation controller

klautcomputing:
The motivation behind his PR is to get rid of the last non websocket requests: Namely when navigating to `/play/game_id` and getting the ELO after a game has ended. This is achieved by implementing a navigation controller which keeps track of the pathname and if the pathname contains a `game_id` it loads the game via the websocket.
While implementing this a couple of other things were fixed/updated:
- resign and draw work again for untimed games
- users are forwarded from the landing page if they join a new game, and are forwarded to the game from anywhere if they join a real time game (no matter whether they accept the game or someone else does). 
- The next game button works again and this time: game disappears on timeout from the next game button and unanswered game controls trigger the next game button too.

### DIFF
--- a/apis/src/components/organisms/display_timer.rs
+++ b/apis/src/components/organisms/display_timer.rs
@@ -1,7 +1,6 @@
 use crate::{
     components::molecules::{live_timer::LiveTimer, user_with_rating::UserWithRating},
-    providers::timer::TimerSignal,
-    responses::user::UserResponse,
+    providers::{auth_context::AuthContext, game_state::GameStateSignal, timer::TimerSignal},
 };
 use hive_lib::color::Color;
 use leptos::*;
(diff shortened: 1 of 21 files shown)


apis/src/components/organisms/display_timer.rs
```diff
@@ -1,31 +1,51 @@
...
+        _ => None,
     };
-    let css_grid_row = match placement {
+    let player_is_black = create_memo(move |_| {
+        user().map_or(false, |user| {
+            let game_state = game_state.signal.get();
+            Some(user.id) == game_state.black_id
+        })
+    });
+    let side = move || match (player_is_black(), placement.unwrap()) {
```
IongIer, line 34: Not safe to unwrap here, this blows up.

PR #143, opened 2024-01-30
Upgrades to leptos 0.6 and leptos-icons 0.3

IongIer:


apis/.rust-analyzer
IongIer, line 1: Not sure about this, didn't fix anything. RA is still completely lost on my machine

klautcomputing (approved):
Clicked around a bit and it seems to work, forwarding after logging in works too. Code looks good as well.

PR #144, opened 2024-02-02
On message refactor

klautcomputing:


apis/src/providers/websocket/challenge/handler.rs
```diff
@@ -0,0 +1,20 @@
+use crate::{common::server_result::ChallengeUpdate, providers::challenges::ChallengeStateSignal};
+
+use leptos::*;
+
+pub fn handle_challenge(challenge: ChallengeUpdate) {
+    match challenge {
+        ChallengeUpdate::Challenges(new_challanges) => {
```
IongIer, line 7: this could use a better name or something, It's the one that handles populating your challenges when you first join and there are a bunch already open?

PR #151, opened 2024-02-18
Uses nested routing for /user/playing and /user/finished

IongIer:


apis/src/pages/profile_view.rs
```diff
@@ -10,8 +12,16 @@ struct UsernameParams {
     username: String,
 }
 
+#[derive(Clone, PartialEq)]
+pub enum ProfileGamesView {
+    Playing,
+    Finished,
+}
+#[derive(Clone)]
+pub struct AllUserGames(pub Vec<GameResponse>, pub Vec<GameResponse>);
```
klautcomputing, line 21: Is the idea here for this to be a placeholder until we make a provider like `Games` for this?

apis/src/pages/display_games.rs
```diff
@@ -0,0 +1,30 @@
+use leptos::*;
+
```
klautcomputing, line 2: Remove empty line and format :)

apis/src/pages/profile_view.rs
```diff
@@ -21,24 +31,74 @@ pub fn ProfileView() -> impl IntoView {
...
         })
     };
-
     let user = Resource::new(username, move |_| get_user_by_username(username()));
+    let games = Resource::new(username, move |_| get_user_games(username()));
+    let stored_children = store_value(children);
+    let button_style = "hover:bg-pillbug-teal transform transition-transform duration-300 active:scale-95 text-white font-bold py-2 px-4 m-1 rounded";
+    let tab_view = create_rw_signal(ProfileGamesView::Playing);
+    let is_playinng = move || {
+        format!(
```
klautcomputing, line 40: Move the if outside and just return to different `String`s?
```
let mut button_style = ...

if matches!(...) {
button_style.push("teal")
} else {
...
}
```

apis/src/pages/profile_view.rs
```diff
@@ -21,24 +31,74 @@ pub fn ProfileView() -> impl IntoView {
...
+            "{button_style} {}",
+            if matches!(tab_view(), ProfileGamesView::Playing) {
+                "bg-pillbug-teal"
+            } else {
+                "bg-ant-blue"
+            },
+        )
+    };
+    let is_finished = move || {
+        format!(
```
klautcomputing, line 50: same here

apis/src/pages/profile_view.rs
klautcomputing: Or is the problem then that this becomes an FnMut?

apis/src/pages/profile_view.rs
```diff
@@ -21,24 +31,74 @@ pub fn ProfileView() -> impl IntoView {
...
+        )
+    };
+    provide_context(tab_view);
 
     view! {
         <div class="bg-light dark:bg-dark pt-12">
             <Transition>
                 {move || {
+                    let maybe_games = games().map(|games| games.ok()).flatten();
+                    let games = match maybe_games {
```
klautcomputing, line 66: I don't think this is required, you get can just do the partition and do

```
let games = {
  maybe_games.sort_by... 
  maybe_games.partition...
}
```
that should yield the same as you match here

klautcomputing (changes requested):
Overall this works :+1: 

But I have a couple of small nitpicks and one bigger question:
Is the idea to turn `AllUserGames` into a provider like `Games` or even merge it with that provider?

PR #183, opened 2024-04-06
Communicate errors

IongIer:


apis/src/pages/register.rs
```diff
@@ -1,68 +1,203 @@
...
                         autocomplete="email"
                         placeholder="Email"
                     />
+                    <p class="h-5">
+                        <Show when=is_invalid_email>
+                            <small class="text-ladybug-red ">"Invalid email"</small>
+                        </Show>
+                    </p>
+                    <small>
+                        We will use it for password reset(when we eventually implement it)
```
klautcomputing, line 144: Notifications and email verification as well?

apis/src/pages/register.rs
```diff
@@ -1,68 +1,203 @@
...
+    let pw_confirm = RwSignal::new(String::new());
+    let has_invalid_char = RwSignal::new(false);
+    let is_invalid_email = RwSignal::new(false);
+    let throttled_input = move |signal_to_update| {
+        use_debounce_fn_with_arg(update_from_input(signal_to_update), 350.0)
+    };
 
+    let pw_invalid = move || {
+        let pw = pw();
+        !(7 < pw.len() && pw == pw_confirm())
```
klautcomputing, line 43: We should check here that it's not greater than 128 chars :)

klautcomputing (changes requested):
Nothing major just some small suggestions

apis/src/pages/register.rs
IongIer: the form prevents you from going over, isn't that enough?

klautcomputing (approved):

PR #191, opened 2024-04-17
Fixes mobile keyboard resizing play area

IongIer:


klautcomputing (approved):
Got 2 folks to test this and they confirmed it works okay on Safari on iOS.

PR #192, opened 2024-04-19
Chat modes

IongIer:


apis/src/websockets/api/game/join_handler.rs
```diff
@@ -18,6 +18,8 @@ use uuid::Uuid;
 pub struct JoinHandler {
     pool: DbPool,
     received_from: actix::Recipient<WsMessage>,
+    // TODO: @ion needs to access to the chat
```
klautcomputing, line 21: remove comments :)

apis/src/websockets/api/game/join_handler.rs
```diff
@@ -63,6 +67,23 @@ impl JoinHandler {
                 username: self.username.to_owned(),
             })),
         });
+        if self.user_id == self.game.white_id || self.user_id == self.game.black_id {
```
klautcomputing, line 70: Can't we do something like this:
```rs
let games = if condition { get private here } else { get public here };
if let Some(messages) =  games.get(...) {
messages.push(...)
}
```

apis/src/websockets/chat.rs
```diff
@@ -0,0 +1,55 @@
...
+            games_public: RwLock::new(HashMap::new()),
+            games_private: RwLock::new(HashMap::new()),
+            direct: RwLock::new(HashMap::new()),
+            direct_lookup: RwLock::new(HashMap::new()),
+        }
+    }
+
+    pub fn insert_or_update_direct_lookup(&self, id1: Uuid, id2: Uuid) {
+        let mut direct_lookup = self.direct_lookup.write().unwrap();
+        let id1_set = direct_lookup.entry(id1).or_insert_with(HashSet::new);
```
klautcomputing, line 49: Does `or_default()` work here?
Which makes me wonder has this been clippied?

apis/src/websockets/chat.rs
```diff
@@ -0,0 +1,55 @@
...
+        id1_set.insert(id2);
+
+        let id2_set = direct_lookup.entry(id2).or_insert_with(HashSet::new);
```
klautcomputing, line 52: Also this can be a one-liner if you want.

apis/src/websockets/internal_server_message.rs
```diff
@@ -11,7 +11,9 @@ pub struct InternalServerMessage {
 #[derive(Debug, Clone)]
 pub enum MessageDestination {
     Direct(actix::Recipient<WsMessage>),
```
klautcomputing, line 13: Should have a comment as well :) IIRC this is not logged in users, right?

apis/src/websockets/lobby.rs
```diff
@@ -281,9 +281,27 @@ impl Handler<ClientActorMessage> for Lobby {
...
+                self.games_users
+                    .entry(game_id.clone())
+                    .or_default()
+                    .insert(cam.from);
+                // Send the message to everyone except white_id and black_id
+                self.games_users
+                    .get(&game_id)
+                    .expect("Game to exists")
+                    .iter()
+                    .for_each(|client| {
```
klautcomputing, line 295: s/client/user/ :)

shared_types/src/chat_message.rs
```diff
@@ -22,6 +30,7 @@ pub struct ChatMessage {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ChatMessageContainer {
     pub destination: ChatDestination,
+    // TODO: @ion maybe even better to change this to messages: Vec<ChatMessage>
```
klautcomputing, line 33: We should leave this in and do this at one point. Because having it send every message individually is actually not great.

klautcomputing (changes requested):

PR #212, opened 2024-05-10
Fixes panic on navigation with chat open and double messages

IongIer:


klautcomputing (approved):

apis/src/components/organisms/chat.rs
```diff
@@ -6,18 +6,28 @@ use leptos::*;
...
 
 #[component]
 pub fn Message(message: ChatMessage) -> impl IntoView {
-    let formatted_timestamp = message
-        .timestamp
-        .unwrap()
-        .format("%Y-%m-%d %H:%M")
-        .to_string();
+    let timestamp = message.timestamp.unwrap().to_string();
+    let options = Intl::DateTimeFormat::new(&Array::new(), &Object::new()).resolved_options();
```
klautcomputing, line 15: This seems complicated to me... let's chat about this.

apis/src/providers/config/configs.rs
```diff
@@ -0,0 +1,28 @@
+use super::confirm_mode::ConfirmMode;
```
klautcomputing, line 1: Why is this completely new?

klautcomputing (changes requested):
Let's figure out a better way to do time

PR #213, opened 2024-05-14
Intersection observer for better panning performance

IongIer:


klautcomputing (approved):
Let's see whether anyone notices a difference.

PR #214, opened 2024-05-15
use reexports to avoid specifying submodules in many cases

PenguinWithATie:
use reexports to avoid specifying submodules in many cases

also some tailwind classes got moved due to my tailwind sorter plugin, I figure it doesn't hurt to have them also but let me know if I should remove them


engine/src/lib.rs
```diff
@@ -1,20 +1,40 @@
...
+mod player;
+mod position;
+mod state;
+mod torus_array;
+mod turn;
+pub use board::Board;
+pub use bug::Bug;
+pub use bug_stack::BugStack;
+pub use color::{Color, ColorChoice};
+pub use dfs_info::DfsInfo;
```
klautcomputing, line 25: `DfsInfo` and `MidMoveBoard` are internal types and it should not be accessible from the outside.
Unsure about the `TorusArray`...

klautcomputing (changes requested):
Thanks for your PR! Looks a lot cleaner now, I just left a comment about 2-3 types that are internal and shouldn't be available from the outside.

engine/src/lib.rs
PenguinWithATie: You are correct, none of these types are required to be public, the Hasher type as well

klautcomputing (approved):
:shipit:

klautcomputing:
Can you make a single commit to be rebased into main, please?

klautcomputing (approved):
Thank you for this!

klautcomputing (approved):
🚢 🇮🇹

PR #218, opened 2024-05-16
fix all clippy warnings

PenguinWithATie:
### Fix all clippy warnings:

- Coment out some fields  in hive_lib that are never read

- Comment out the new server message function as it is never used

- Make the ChalllengeDetails shared type to improve argument passing on Challenge creation

- Put the Game variant of ServerMessage and the Ok variant of serverResult in a Box pointer (as per the analyzer's suggestion to improve performance)


klautcomputing (changes requested):
This is a great clean up PR. Thank you 👍 
Only one thing, I would go ahead and remove the code you currently only commented out.

IongIer (approved):

klautcomputing (approved):
Thank you!

PR #223, opened 2024-05-19
add "quickplay" feature

PenguinWithATie:
## Created the quickplay feature

- Create grid of buttons with predetermined game settings
- Simple auto match, if a user tries to create a game with settings compatible with an existing game, it will join that instead. 
- Block creating games with identical settings (so if a user spams the same button only a single game will be created) for real time games only
- Add the play button into the grid as "custom"

### Notes
 - the styling of the buttons may be improved later

### DIFF
--- a/apis/src/components/organisms/quickplay.rs
+++ b/apis/src/components/organisms/quickplay.rs
@@ -1,0 +1,120 @@
+use crate::{
+    common::ChallengeAction,
+    components::{atoms::rating::icon_for_speed, molecules::modal::Modal},
+    pages::challenge_create::ChallengeCreate,
+    providers::{ApiRequests, AuthContext},
+};
+use core::panic;
+use hive_lib::{ColorChoice, GameType};
+use leptos::{html::Dialog, *};
+use leptos_icons::*;
+use leptos_router::use_navigate;
+use shared_types::ChallengeVisibility;
+use shared_types::{ChallengeDetails, GameSpeed::*, TimeMode};
+pub enum QuickPlayTimeControl {
+    Bullet1p2,
+    Blitz3p3,
+    Blitz5p4,
+    Rapid10p10,
+    Rapid15p10,
+    Classic20p20,
+    Classic30p30,
+}
+use QuickPlayTimeControl::*;
+const BUTTON_STYLE: &str = "flex justify-center px-4 py-2 font-bold text-white rounded bg-button-dawn dark:bg-button-twilight hover:bg-pillbug-teal active:scale-95";
+
+#[component]
+pub fn GridButton(time_control: QuickPlayTimeControl) -> impl IntoView {
+    let (display_text, icon_data, base, increment) = match time_control {
+        Bullet1p2 => ("1+2".to_owned(), icon_for_speed(&Bullet), 1, 2),
+        Blitz3p3 => ("3+3".to_owned(), icon_for_speed(&Blitz), 3, 3),
+        Blitz5p4 => ("5+4".to_owned(), icon_for_speed(&Blitz), 5, 4),
+        Rapid10p10 => ("10+10".to_owned(), icon_for_speed(&Rapid), 10, 10),
+        Rapid15p10 => ("15+10".to_owned(), icon_for_speed(&Rapid), 15, 10),
+        Classic20p20 => ("20+20".to_owned(), icon_for_speed(&Classic), 20, 20),
+        Classic30p30 => ("30+30".to_owned(), icon_for_speed(&Classic), 30, 30),
+    };
+    view! {
+        <button
+            class=BUTTON_STYLE
+            on:click=move |_| {
+                let auth_context = expect_context::<AuthContext>();
+                let account = match (auth_context.user)() {
+                    Some(Ok(Some(account))) => Some(account),
+                    _ => None,
+                };
+                if account.is_some() {
+                    let api = ApiRequests::new();
+                    let details = ChallengeDetails {
+                        rated: true,
+                        game_type: GameType::MLP,
+                        visibility: ChallengeVisibility::Public,
+                        opponent: None,
+                        color_choice: ColorChoice::Random,
+                        time_mode: TimeMode::RealTime,
+                        time_base: Some(base * 60),
+                        time_increment: Some(increment),
+                        band_upper: None,
+                        band_lower: None,
+                    };
+                    let challenge_action = ChallengeAction::Create(details);
+                    api.challenge(challenge_action);
+                } else {
+                    let navigate = use_navigate();
+                    navigate("/login", Default::default());
+                }
+            }
+        >
+
+            <Icon icon=icon_data/>
+            {display_text}
+        </button>
+    }
+}
+#[component]
+pub fn QuickPlay() -> impl IntoView {
+    let open = create_rw_signal(false);
+    let dialog_el = create_node_ref::<Dialog>();
+    let close_modal = Callback::new(move |()| {
+        dialog_el
+            .get_untracked()
+            .expect("dialog to have been created")
+            .close();
+    });
+    view! {
+        <div>
+            <Modal open=open dialog_el=dialog_el>
+                <ChallengeCreate close=close_modal/>
+            </Modal>
+            <span class="flex justify-center mb-4 text-xl font-bold">Quick Play</span>
+            <div class="grid grid-cols-2 gap-4 w-full md:grid-cols-3">
+                <GridButton time_control=Bullet1p2/>
+                <GridButton time_control=Blitz3p3/>
+                <GridButton time_control=Blitz5p4/>
+                <GridButton time_control=Rapid10p10/>
+                <GridButton time_control=Rapid15p10/>
+                <GridButton time_control=Classic20p20/>
+                <GridButton time_control=Classic30p30/>
+                <button
+                    class=BUTTON_STYLE
+                    on:click=move |_| {
+                        let auth_context = expect_context::<AuthContext>();
+                        let account = match (auth_context.user)() {
+                            Some(Ok(Some(account))) => Some(account),
+                            _ => None,
+                        };
+                        if account.is_some() {
+                            open.update(move |b| *b = true)
+                        } else {
+                            let navigate = use_navigate();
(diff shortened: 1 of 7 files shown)


apis/src/components/organisms/quickplay.rs
```diff
@@ -0,0 +1,83 @@
...
+                    "5+0" => (5, 0),
+                    "5+3" => (5, 3),
+                    "10+0" => (10, 0),
+                    "10+5" => (10, 5),
+                    "15+10" => (15, 10),
+                    _ => panic!("Invalid text"),
+                };
+                if account.is_some() {
+                    let details = ChallengeDetails {
+                        rated: false,
```
IongIer, line 41: Rated needs to be true here.

apis/src/components/organisms/quickplay.rs
```diff
@@ -0,0 +1,83 @@
...
+use crate::{
+    common::ChallengeAction,
+    providers::{ApiRequests, AuthContext},
+};
+use hive_lib::{ColorChoice, GameType};
+use shared_types::ChallengeVisibility;
+use shared_types::{ChallengeDetails, TimeMode};
+
+#[component]
+fn GridButton(text: String) -> impl IntoView {
```
IongIer, line 15: Knowing leex I suspect he'll want to use an enum here instead of text: String, or something like default_speed: Type.

apis/src/components/organisms/quickplay.rs
```diff
@@ -0,0 +1,83 @@
...
+                    "2+1" => (2, 1),
+                    "3+0" => (3, 0),
+                    "3+2" => (3, 2),
+                    "5+0" => (5, 0),
+                    "5+3" => (5, 3),
+                    "10+0" => (10, 0),
+                    "10+5" => (10, 5),
+                    "15+10" => (15, 10),
+                    _ => panic!("Invalid text"),
+                };
```
IongIer, line 38: This could be moved into the account.is_some part imo.

apis/src/components/organisms/quickplay.rs
```diff
@@ -0,0 +1,83 @@
...
+            {display_text}
+        </button>
+    }
+}
+#[component]
+pub fn QuickPlay() -> impl IntoView {
+    view! {
+        <div>
+            <span class="text-xl font-bold">Quick Play</span>
+            <div class="flex grid grid-cols-2 gap-4 w-full md:grid-cols-3">
```
IongIer, line 70: flex and grid apply the same property, and it can lead to weird behavior between browsers because they apply the classes slightly differently. Just grid here

apis/src/components/organisms/quickplay.rs
```diff
@@ -0,0 +1,83 @@
...
+                <GridButton text="1+0".to_owned()/>
+                <GridButton text="2+1".to_owned()/>
+                <GridButton text="3+0".to_owned()/>
+                <GridButton text="3+2".to_owned()/>
+                <GridButton text="5+0".to_owned()/>
+                <GridButton text="5+3".to_owned()/>
+                <GridButton text="10+0".to_owned()/>
+                <GridButton text="10+5".to_owned()/>
+                <GridButton text="15+10".to_owned()/>
```
IongIer, line 79: 1+2, 3+3, 5+4(5+5 is already rapid and I think two blitz options is better), 10+10, 20+20, and a Classical one...30+30?

apis/src/providers/api_requests.rs
```diff
@@ -77,9 +78,25 @@ impl ApiRequests {
...
-        self.websocket
-            .send(&serde_json::to_string(&msg).expect("Serde_json::to_string failed"));
+        let challenge_action = match challenge_action {
+            ChallengeAction::Create(details) => {
+                let auth_context = expect_context::<AuthContext>();
+                let account = match (auth_context.user)() {
+                    Some(Ok(Some(account))) => Some(account),
+                    _ => None,
+                };
+                let user = account.unwrap().user.username;
```
IongIer, line 88: unwrapping here shouldn't panic but just to be safe an if let Some(username) might be a better way to go

apis/src/responses/challenge.rs
```diff
@@ -70,5 +72,77 @@ impl ChallengeResponse {
...
+    details.game_type == GameType::from_str(&challenge.game_type).unwrap()
+        && details.rated == challenge.rated
+        && details.time_mode == challenge.time_mode
+        && details.time_base == challenge.time_base
+        && details.time_increment == challenge.time_increment
+        && match details.color_choice {
+            ColorChoice::Random => true,
+            ColorChoice::White => {
+                challenge.color_choice == ColorChoice::Black
+                    || challenge.color_choice == ColorChoice::Random
```
IongIer, line 90: Don't think we should match people looking for Random with people looking for a certain color. I might be fine playing Random but wouldn't accept a challenge where it's certain I'll play Black for example

apis/src/responses/challenge.rs
```diff
@@ -70,5 +72,77 @@ impl ChallengeResponse {
...
+        && if let Some(opponent) = &details.opponent {
+            opponent == &challenge.challenger.username
+        } else if let Some(opponent) = &challenge.opponent {
+            opponent.username == user
+        } else {
+            challenge.challenger.username != user
+        }
+}
+
+fn has_same_details(challenge: &ChallengeResponse, user: &str, details: &ChallengeDetails) -> bool {
```
IongIer, line 106: Something like this might be good thing to add but only for realtime games. Creating two games with identical settings for correspondence shouldn't be stopped. Maybe I want to start playing correspondence more seriously I should be allowed to open many challenges at once.

apis/src/components/organisms/quickplay.rs
PenguinWithATie: made rated true but also set the ELO upper bound to +300 since auto matching will happen. let me know if this should be adjusted

apis/src/responses/challenge.rs
PenguinWithATie: Good point, unlimited time games and correspondence should not have this restriction

apis/src/components/organisms/quickplay.rs
IongIer: Leave it None for now, otherwise lgtm. Let's see what leex says

apis/src/responses/challenge.rs
```diff
@@ -70,5 +72,71 @@ impl ChallengeResponse {
...
+    challenges: Values<String, ChallengeResponse>,
+) -> Option<ChallengeAction> {
+    let mut matching_id = None;
+    let mut exists = false;
+    for challenge in challenges {
+        if is_compatible(challenge, &user, &details) {
+            matching_id = Some(challenge.nanoid.clone());
+            break;
+        }
+        if details.time_mode == TimeMode::RealTime && has_same_details(challenge, &user, &details) {
```
klautcomputing, line 130: replace with https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any

apis/src/responses/challenge.rs
```diff
@@ -70,5 +72,71 @@ impl ChallengeResponse {
...
+
+pub fn create_challenge_handler(
+    user: String,
+    details: ChallengeDetails,
+    challenges: Values<String, ChallengeResponse>,
+) -> Option<ChallengeAction> {
+    let mut matching_id = None;
+    let mut exists = false;
+    for challenge in challenges {
+        if is_compatible(challenge, &user, &details) {
```
klautcomputing, line 126: replace this with a https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find, while what I am suggesting is minimally slower it makes your code a lot more readable, because you don't need to initialize 2 `mut` variables and there is no need for `break`.

klautcomputing (changes requested):

apis/src/providers/api_requests.rs
```diff
@@ -77,9 +78,28 @@ impl ApiRequests {
...
+                let auth_context = expect_context::<AuthContext>();
+                let account = match (auth_context.user)() {
+                    Some(Ok(Some(account))) => Some(account),
+                    _ => None,
+                };
+                if let Some(account) = account {
+                    let challenges = expect_context::<ChallengeStateSignal>().signal.get();
+                    let challenges = challenges.challenges.values();
+                    create_challenge_handler(account.user.username, details, challenges)
+                } else {
```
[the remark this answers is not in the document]
PenguinWithATie: an else is required, the default return type is "unit" () not the None variant

apis/src/responses/challenge.rs
```diff
@@ -70,5 +72,71 @@ impl ChallengeResponse {
...
+}
+}
+
+fn is_compatible(challenge: &ChallengeResponse, user: &str, details: &ChallengeDetails) -> bool {
+    details.game_type == GameType::from_str(&challenge.game_type).unwrap()
+        && details.rated == challenge.rated
+        && details.time_mode == challenge.time_mode
+        && details.time_base == challenge.time_base
+        && details.time_increment == challenge.time_increment
+        && match details.color_choice {
```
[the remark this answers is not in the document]
PenguinWithATie: When its random, that is fine, but when its not, oposite colors must be matched (and as per @IongIer 's suggestion, it should not match e.g white with random)

IongIer (approved):

apis/src/responses/challenge.rs
```diff
@@ -70,5 +72,61 @@ impl ChallengeResponse {
...
+        && challenge_opponent == details.opponent.as_deref()
+        && challenge.challenger.username == user
+}
+
+pub fn create_challenge_handler(
+    user: String,
+    details: ChallengeDetails,
+    mut challenges: Values<String, ChallengeResponse>,
+) -> Option<ChallengeAction> {
+    if details.time_mode == TimeMode::RealTime
```
klautcomputing, line 123: I would have done an early return, but that's fine too.
Just in case I don't make sense:
```rs
if ... {
return None;
}
if let ... {
return Some(...);
}
Some(...)
```

klautcomputing (approved):
I left one nitpick comment you can ignore if you want. Thanks for addressing all my feedback! :)

PR #224, opened 2024-05-19
Adds admins with superpowers

IongIer:
A simple "admin panel" that allows sending global messages.

Bonus: Link to the official pdf rules.


db/migrations/2024-05-19-135658_make_admins/up.sql
```diff
@@ -0,0 +1 @@
+alter table users add column admin bool not null default false;
```
klautcomputing, line 1: newline please :)

apis/src/websockets/api/chat/handler.rs
```diff
@@ -70,6 +70,10 @@ impl ChatHandler {
                     message: ServerMessage::Chat(vec![self.message.to_owned()]),
                 })
             }
+            ChatDestination::Global => messages.push(InternalServerMessage {
```
klautcomputing, line 73: This needs a check that the user who's sending the global message is actually allowed to do so, so please check that they are an admin, here.

klautcomputing (changes requested):

klautcomputing (approved):
That should do the trick :)

PR #235, opened 2024-05-24
Refactors game_state, adds finer grained reactivity based on slices and batches updates

IongIer:
This is more of an experiment where I try to make our re renders finer grained and hopefully fix some of our performance issues on older phones. A work in progress, excuse the mess :)

This sort of refactor could come with drawbacks according to [The Scriptures](https://book.leptos.dev/15_global_state.html#option-3-create-a-global-state-struct-and-slices)

Take a look at what I'm trying to do, feedback appreciated.


PenguinWithATie:
For better readability I would encapsulate all the slice creation in a trait and then call from there, for example:

```
trait GameSignalSlices {
    fn  get_state_slice() -> Signal<State> 
}
```

PenguinWithATie (approved):

klautcomputing (approved):

PR #244, opened 2024-06-12
Avoids opening multiple pools per request

klautcomputing:
This change gets the code base away from making multiple pools all the time.
It now has one (1) pool per request and every DB interaction for that request is on the same connection.
Now `Handler`s get a `pool`, make a `connection` and then from everything in the models uses a `DbConn` instead of a `DbPool`
`Handler`s are now responsible for opening a `transaction` if they are doing writes to multiple `models`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -424,6 +424,7 @@ dependencies = [
  "cookie 0.18.1",
  "db",
  "diesel",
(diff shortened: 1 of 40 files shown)


PenguinWithATie (approved):
Looks great, awsome to get this one in

PR #270, opened 2024-07-03
analysis optimization

PenguinWithATie:
this should speed up analysis since insertions to the hashmap (hence clones) are greatly reduced, should not matter much for smaller trees (which is most trees) but nice to have


klautcomputing (commented):
Either the branches are empty and then we don't need the assert or the assertion blows up, so I think it doesn't make sense to have it in there. If this is for debugging purposes you can change it to a debug assert.

klautcomputing (approved):

PR #281, opened 2024-07-09
analysis improvements

PenguinWithATie:
If a state has been reached previously it willl jump to that tree in the node if the turn numbers match.

If at the current node a 3 move rep has ocurred mark all corresponding turns with the rep symbol

A bijective map is needed instead of regular hashmap since we need to search by both sides (by "key" and "value") for insert and delete respectivly

With  tree_ds 0.15, we can use the [compact_serde](https://github.com/clementwanjau/tree-ds/issues/12) feature and the size of the file will scale with the number of nodes, not the complexity of the tree. The size seems to be 1kb per 10 nodes including the BiMap. So we use standard json serialization for better human readability and simple interop with future systems.

Avoid repeated ids in the case of deleting a node

### DIFF
--- a/apis/src/components/atoms/target.rs
+++ b/apis/src/components/atoms/target.rs
@@ -30,13 +30,16 @@ pub fn Target(
                 }
                 analysis.update(|analysis| {
                     if let Some(analysis) = analysis {
-                        let moves = game_state.signal.get_untracked().state.history.moves;
-                        let last_move = moves.last().unwrap().clone();
-                        if last_move.0 == "pass" {
+                        let state = game_state.signal.get_untracked().state;
+                        let moves = state.history.moves;
+                        let hashes = state.hashes;
+                        let last_index = moves.len() - 1;
+                        if moves[last_index].0 == "pass" {
                             //if move is pass, add prev move
-                            analysis.add_node(moves[moves.len() - 2].clone());
+                            analysis
+                                .add_node(moves[last_index - 1].clone(), hashes[last_index - 1]);
                         }
-                        analysis.add_node(last_move);
+                        analysis.add_node(moves[last_index].clone(), hashes[last_index]);
                     }
                 });
             });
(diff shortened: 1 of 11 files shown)


apis/src/components/atoms/target.rs
```diff
@@ -30,13 +30,16 @@ pub fn Target(
...
-                        if last_move.0 == "pass" {
+                        let state = game_state.signal.get_untracked().state;
+                        let moves = state.history.moves;
+                        let hashes = state.hashes;
+                        let last_index = moves.len() - 1;
+                        if moves[last_index].0 == "pass" {
                             //if move is pass, add prev move
-                            analysis.add_node(moves[moves.len() - 2].clone());
+                            analysis
+                                .add_node(moves[last_index - 1].clone(), hashes[last_index - 1]);
```
klautcomputing, line 40: I am not sure this is correct, because "passes" have their own individual hash which is changing to reflect which piece is not stunned anymore, which side is to play, etc.

apis/src/components/organisms/analysis/atoms.rs
```diff
@@ -27,6 +27,14 @@ pub fn UndoButton() -> impl IntoView {
                     let new_current = node.get_parent_id();
                     if let Some(new_current) = new_current {
                         a.update_node(new_current);
+                        a.tree
+                            .get_subtree(&node.get_node_id(), None)
+                            .unwrap()
```
klautcomputing, line 32: Can we do a `if let Some/Ok()` here please

klautcomputing (changes requested):
So this unfortunately doesn't work, the way you use the has is too restrictive. We deployed it to hivegame.fly.dev and I tried a couple of things: play 2 queens and then walk the queens around a bit, it doesn't show the games has finished by 3 move repetition. 
Or try and play the same opener but rotated, so let's say wL, bL, wQ, bQ in a Z and add an ant to you queen then go back to the start and play the same Z but rotated or mirrored and spawn another ant. And you will get to a broken state.
This is because you use the hash to get to a position, but the string of moves does not make sense now anymore when navigating there.
Maybe I am not explaining this perfectly... I can show you if we get on a call :)
I attached some screenshots to confuse you even more. 

Screenshot 2024-07-09 at 10 29 03
Screenshot 2024-07-09 at 10 24 27
Screenshot 2024-07-09 at 10 26 25

IongIer (approved):

IongIer (approved):

klautcomputing (approved):

IongIer (approved):

PR #291, opened 2024-07-13
integrate more leptix components

PenguinWithATie:
-Use radio group to simpify create_challenge (works and looks the same) 

-remove dropdows from time_select and use radio groups (looks better also imo)

-Use leptix switches instead of checkboxes on tournament page for a more uniform feel

-Progress bar for tournaments


klautcomputing (changes requested):
I left another small comment and a screenshot on discord :)

PR #294, opened 2024-07-18
Tournament Schedules

PenguinWithATie:
Adds scheduling for games, right now only works with tournaments but could be integrated elsewhere

Update: Allows multiple schedules per game, some refactoring

TODO: Integrate with notifications

Notes: In the current implementation any player in a game can reject a confirmed or unconfirmed date without any additional checks, so this should only be used for the convenience of users but not to enforce any rules.


klautcomputing (changes requested):
I mainly looked at the backend side of things, because that's where I start from when reviewing code and left you a couple of comments.

Happy to discuss how we can move this forward on discord or here 😄

db/migrations/2024-07-16-021141_add_game_start_date/down.sql
```diff
@@ -0,0 +1,3 @@
+alter table games drop column start_date;
```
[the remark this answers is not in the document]
PenguinWithATie: Agreed I will change the implementation

apis/src/websockets/api/game/handler.rs
```diff
@@ -112,6 +113,18 @@ impl GameActionHandler {
                 .handle()
                 .await?
             }
+            GameAction::SetDate(date) => {
+                self.ensure_user_is_player()?;
+                UpdateDateHandler::new(&self.game, self.user_id, &self.pool, date)
+                    .handle()
+                    .await?
+            }
+            GameAction::SetAcceptDate(accept, white_proposer) => {
```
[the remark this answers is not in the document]
PenguinWithATie: this is correct and this is a "bug" but i think this isint a bug deal as per the note, because anyway at any point any of the players can delete a date and there is no check for this (and no notification or interaction with the tourlament organizer for that matter). If we will use this feature to enforce rules, I will need a closer specification of what is needed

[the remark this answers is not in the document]
PenguinWithATie: Also as of now if a schedule expires any player is able to simply schedule again. If we will make schedules more "binding" it would be good to know what is expected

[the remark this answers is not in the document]
klautcomputing: What I was trying to say is: The backend should never rely on the user feeding it accurate data. We cannot rely on data the user sends, especially when it gives one user the option to make a decision for another user.
The data we can trust is the `user` we extract from the cookie and we can then trust the `accept/decline` that was delivered via the connection that is authenticated by the cookie. And the color of the user and the proposer needs to be established from the database.

apis/src/websockets/api/schedules/handler.rs
```diff
@@ -0,0 +1,297 @@
...
+        let mut messages = Vec::new();
+        let mut conn = get_conn(&self.pool).await?;
+        match &self.action {
+            ScheduleAction::Propose(date, id) => {
+                let opponent_id = conn
+                    .transaction::<_, anyhow::Error, _>(move |tc| {
+                        async move {
+                            let game = Game::find_by_game_id(id, tc).await?;
+                            self.ensure_user_is_player(&game)?;
+                            diesel::insert_into(game_schedules::table)
```
klautcomputing, line 51: Please move this functionality into its respective model. Doing it here mixes concerns and doesn't make the model reusable. 
Same for the other places.

apis/src/websockets/api/schedules/handler.rs
```diff
@@ -0,0 +1,297 @@
+use std::collections::HashMap;
+
```
klautcomputing, line 2: remove line

apis/src/websockets/api/schedules/handler.rs
```diff
@@ -0,0 +1,297 @@
...
+use db_lib::{
+    get_conn,
+    models::{Game, Tournament},
+    schema::{
+        game_schedules::{self},
+        games::{self},
+    },
+    DbPool,
+};
+
```
klautcomputing, line 19: same here

db/migrations/2024-07-16-021141_add_game_start_date/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE game_schedules (
+  game_id uuid references games(id) on delete cascade,
+  proposer_id uuid references users(id) on delete cascade,
+  start_t TIMESTAMP WITH TIME ZONE,
+  end_t TIMESTAMP WITH TIME ZONE,
+  agreed BOOLEAN NOT NULL DEFAULT false,
+  primary key(game_id, proposer_id)
```
klautcomputing, line 7: Doesn't this you can only propose one date?

apis/src/providers/schedules.rs
```diff
@@ -0,0 +1,16 @@
+use std::collections::HashMap;
+
+use leptos::{provide_context, RwSignal};
+use shared_types::GameId;
+
+use crate::responses::ScheduleResponse;
+
+#[derive(Default, Clone)]
+pub struct SchedulesContext {
+    pub my_schedules: RwSignal<HashMap<GameId, Vec<ScheduleResponse>>>,
```
klautcomputing, line 10: What does splitting them into two help with?

apis/src/websockets/api/schedules/handler.rs
```diff
@@ -0,0 +1,297 @@
...
+                    destination: MessageDestination::User(self.user_id),
+                    message: ServerMessage::Schedule(ScheduleUpdate::OwnTournamentSchedules(
+                        schedules,
+                    )),
+                });
+            }
+        }
+        Ok(messages)
+    }
+    fn ensure_user_is_player(&self, game: &Game) -> Result<()> {
```
klautcomputing, line 288: This could maybe go into the request handler? So that we can share it with other handlers which do the same check. Not 100% sure though.

klautcomputing (changes requested):
There is a couple of things that I think should be changed:
1. TournamentResponse should get the schedules added to it.
2. The model needs to have CRUD methods not the handler.
3. We might want to allow people to propose more than one date per game?
4. The tournament page needs to be reactive.

I left a bunch of comments in the relevant places.

apis/src/components/atoms/schedule_controls.rs
```diff
@@ -0,0 +1,90 @@
...
+            <Show when=move || is_expired_or_none>
+                <DateTimePicker
+                    text=""
+                    min=Local::now() + Duration::minutes(10)
+                    max=Local::now() + Duration::weeks(12)
+                    success_callback=Callback::from(move |utc| {
+                        selected_time.set(utc);
+                    })
+
+                    failure_callback=Callback::from(|_| { logging::log!("Failure to set date") })
```
IongIer, line 50: I would get rid of these logging callbacks, we can make the failure callback optional on <DateTimePicker/> and then it can be called with or without a failure one.

apis/src/pages/tournament.rs
```diff
@@ -362,3 +473,83 @@ pub fn Tournament() -> impl IntoView {
         </div>
     }
 }
+
+#[component]
+pub fn MySchedules(
+    games_hashmap: Memo<HashMap<GameId, GameResponse>>,
+    user_id: Option<Uuid>,
+) -> impl IntoView {
+    logging::log!("{:?}", user_id);
```
IongIer, line 482: As soon as we know it works correctly I would get rid of these logs/ maybe just comment them out

apis/src/pages/tournament.rs
```diff
@@ -40,6 +48,13 @@ pub fn Tournament() -> impl IntoView {
         Some(Ok(Some(account))) => Some(account),
         _ => None,
     };
+    create_effect(move |_| {
```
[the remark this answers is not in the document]
IongIer: I can fix this part if you want along with any styling that needs changing after the backend stuff is addressed.

apis/src/providers/schedules.rs
PenguinWithATie: a logged off user should still be able to see all agreed schedules hence its safer to isolate the info sent to each client so it will only contain schedules relevant to them

db/src/models/schedule.rs
```diff
@@ -0,0 +1,155 @@
...
+            Err(DbError::Unauthorized)
+        }
+    }
+
+    pub async fn get_first_agreed(game_id: Uuid, conn: &mut DbConn<'_>) -> Result<Self, DbError> {
+        Ok(game_schedules::table
+            .filter(game_schedules::game_id.eq(game_id))
+            .filter(game_schedules::agreed.eq(true))
+            .filter(game_schedules::start_t.gt(now))
+            .first(conn)
```
klautcomputing, line 102: Is this deterministic? Will we get the same results if we run this function twice? I actually don't know...

klautcomputing (changes requested):
Sorry, I should have been a little more precise when it came to how to move this code into the model, because most my feedback revolves around that.
But if we address that (and I am happy to help here!) I think we can merge that.
I really like the functionality and how it works now 👍

db/src/models/schedule.rs
PenguinWithATie: Likely not totally deterministic, im trying to avoid sending expired schedules but there is also filtering on the client side so this can be removed if you want

db/migrations/2024-07-16-021141_add_game_start_date/up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE game_schedules (
+  id uuid default gen_random_uuid() primary key not null,
+  game_id uuid references games(id) on delete cascade not null,
+  proposer_id uuid references users(id) on delete cascade not null,
+  start_t TIMESTAMP WITH TIME ZONE not null,
```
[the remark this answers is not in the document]
PenguinWithATie: I see finding the other player/game as a form or model validation. but im fine with either appoach

db/src/models/schedule.rs
```diff
@@ -0,0 +1,155 @@
...
+            Ok(GameId(game.nanoid))
+        } else {
+            Err(DbError::Unauthorized)
+        }
+    }
+
+    pub async fn insert(&self, conn: &mut DbConn<'_>) -> Result<usize, DbError> {
+        Ok(diesel::insert_into(game_schedules::table)
+            .values((
+                game_schedules::id.eq(self.id),
```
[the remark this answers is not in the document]
PenguinWithATie: yes, this was a hack, now tthat I understand the types better I have fixed

klautcomputing (changes requested):
A couple of nitpicks that can you can do but no need and one real issue: the primary key 🙂

db/migrations/2024-07-16-021141_add_game_start_date/up.sql
IongIer, line 1: The folder here should be renamed to something more descriptive I think. add_schedules or something like that.

IongIer (approved):

klautcomputing (approved):
TLDR 👍

PR #295, opened 2024-07-18
Server side ping

klautcomputing:
Instead of having the user determine what their ping is, we need a server side ping. So the server pings the user and they respond. And the server knows the ping.
Then we can use the ping which we now know on the server side to do lag compensation.


klautcomputing:
I can't computer today, it's too hot. 😅 So I am gonna give myself a little more time for this.

IongIer (approved):

IongIer (approved):

PR #297, opened 2024-07-20
Lag compensation

klautcomputing:


apis/src/lag_tracking/lag_tracker.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    pub fn comp_est_std_err(&self) -> Option<f64> {
+        if self.moves() > 2 {
+            Some(self.comp_est_sq_err.sqrt() / ((self.moves() - 2) as f64))
+        } else {
+            None
+        }
+    }
```
nrobinaubertin, line 82: I don't think that this is used anywhere

apis/src/lag_tracking/lag_tracker.rs
```diff
@@ -0,0 +1,103 @@
...
+#[derive(Clone, Debug)]
+pub struct LagTracker {
+    pub quota_gain: f64,
+    pub quota: f64,
+    pub quota_max: f64,
+    pub lag_estimator: DecayingStats,
+    pub uncomp_stats: Stats,
+    pub lag_stats: Stats,
+    pub comp_est_sq_err: f64,
+    pub comp_est_overs: f64,
```
nrobinaubertin, line 12: I think that those two values are not really used to calculate the compensation in the end

nrobinaubertin (commented):
Some leftovers from the lichess code translation but LGTM overall

klautcomputing:
@nrobinaubertin True but I thought I might keep them around because I want to make a statistics page where you can see all the numbers which might in the future help us fine tune the lag compensation :)

IongIer (approved):

PR #309, opened 2024-07-28
fix analysis from history

PenguinWithATie:
Fixes a bug where analysis doesnt load the game correctly if entering from a point it history other than the last (currently only possible on mobile)


IongIer:
This is a great change! I think we should also support doing this in the desktop view as well. 

We can merge this as is or if you want to you can add the download/analysis buttons in the desktop history as well.

IongIer (approved):

klautcomputing (approved):
Thank you!

PR #312, opened 2024-07-30
Updates to leptos use 0.11, changes to msgpack encoding

IongIer:
Upgrades to leptos_use 0.11, this allows us to use codecs for our ws messages. 

Wip: Need to use better types on the ws messages, the current changes are just following compiler suggestions to get it working again after the breaking changes introduced by leptos_use

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -354,7 +354,7 @@ dependencies = [
  "getrandom",
  "once_cell",
  "version_check",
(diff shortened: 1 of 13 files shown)


PenguinWithATie:
are you using strings encoding rather than binary for any specific reason?

klautcomputing (approved):
Thanks for getting rid of the expects!

PR #314, opened 2024-07-31
use messagepack codec

PenguinWithATie:
This change is useless on its own. The main point is to show that using a binary encoding works correctly.

Also, once the String type on the socket is replaced by some nicer type (and hence the underlying conversion to and from JSON is avoided) messages would become substantially smaller

### DIFF
--- a/apis/src/websockets/ws_connection.rs
+++ b/apis/src/websockets/ws_connection.rs
@@ -1,6 +1,6 @@
 use super::tournament_game_start::TournamentGameStart;
 use super::{api::request_handler::RequestHandler, internal_server_message::MessageDestination};
-use crate::common::{ClientRequest, ExternalServerError, ServerResult};
+use crate::common::{CommonMessage, ExternalServerError, ServerResult};
 use crate::lag_tracking::lags::Lags;
 use crate::ping::pings::Pings;
 use crate::websockets::{
@@ -14,11 +14,11 @@ use actix::{
 };
 use actix_web_actors::ws::{self};
 use anyhow::Result;
+use codee::{binary::MsgpackSerdeCodec, Decoder, Encoder};
 use db_lib::DbPool;
 use shared_types::SimpleUser;
 use std::time::{Duration, Instant};
 use uuid::Uuid;
-use codee::{binary::MsgpackSerdeCodec, Decoder, Encoder};
 
 const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
 const CLIENT_TIMEOUT: Duration = Duration::from_secs(10);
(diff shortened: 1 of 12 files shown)


apis/src/websockets/ws_connection.rs
```diff
@@ -135,8 +136,10 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for WsConnection {
                 ctx.stop();
             }
             Ok(ws::Message::Nop) => (),
-            Ok(Text(s)) => {
-                if let Ok(request) = serde_json::from_str::<ClientRequest>(s.as_ref()) {
+            Ok(ws::Message::Binary(s)) => {
+                let request: Result<String,_> = MsgpackSerdeCodec::decode(&s);
+                let request = serde_json::from_str::<ClientRequest>(request.unwrap().as_ref());
```
klautcomputing, line 141: 1. We can't hard unwrap here, we need to do an `if let Some` or something like that.
2. Why are we double encoding the messages now? First as JSON and then on top of that as MsgPack. That doesn't make sense to me, we should only encode them once. 

Why I am so strict here is because people will still have the old wasm running on their devices and they will still send with the old code and it will fail the unwrap here.

apis/src/websockets/ws_connection.rs
```diff
@@ -172,6 +175,8 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for WsConnection {
                                         Box::new(message.message),
                                     ))
                                     .expect("Failed to serialize a server message");
```
klautcomputing, line 177: Reviewing my own code here... All the expects should be removed.

apis/src/websockets/ws_connection.rs
```diff
@@ -172,6 +175,8 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for WsConnection {
...
+                                    let serialized = MsgpackSerdeCodec::encode(&serialized)
```
klautcomputing, line 178: And no new ones added 🙂

apis/src/websockets/ws_connection.rs
```diff
@@ -198,6 +203,8 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for WsConnection {
                                 });
                                 let serialized = serde_json::to_string(&message)
                                     .expect("Failed to serialize a server message");
+                                let serialized = MsgpackSerdeCodec::encode(&serialized)
```
klautcomputing, line 206: Same here

apis/src/websockets/ws_server.rs
```diff
@@ -80,6 +82,8 @@ impl Handler<Ping> for WsServer {
             }));
             let serialized =
                 serde_json::to_string(&message).expect("Failed to serialize a server message");
+            let serialized = MsgpackSerdeCodec::encode(&serialized)
```
klautcomputing, line 85: Same comment as above for the whole file 🙂

klautcomputing (changes requested):
I think it would make sense to not double encode and we should get rid of all expects from the WebSocket code, both on the server and client side.

PenguinWithATie:
> I think it would make sense to not double encode and we should get rid of all expects from the WebSocket code, both on the server and client side.

Now with `CommonMessage` there is no more double encoding, though im sure the types could be reworked

IongIer (approved):

PR #315, opened 2024-07-31
Refactors and styles tournament page and refactors schedules

IongIer:


PenguinWithATie:
This looks amazing thank you!!!!

Couple of small suggestions/nitpicks (not needed imo but would be nice)

-Hide Pending games and MySchedules if there is no content no display
-Display previews in a normal grid (and not a column) when not on mobile

PR #318, opened 2024-08-02
extend game search api

PenguinWithATie:
#### Extend api to search games in a more generic way, It requires to pass an option to select the context to update, (for now only signals related to profile games).

- Remove the server functions to get profile games and use the ws api, this allows for a substantial refactor on profile view and no longer requires the use of resources.

- Remove the server function to get user detalis and do it through ws also

- In the Game model of db, get_ongoing_games_for_username and get_x_finished_games_for_username are now both done by get_rows_from_options

- Allows filtering by speed on profile

- Allows filtering by color on profile

- Allows multiple users (currently not used in FE)

- Allows filtering by win/loss/draw

#### Minor fixes
- Slight sound rework of sounds, avoid the use of [use_timeout_fn](https://leptos-use.rs/animation/use_timeout_fn.html#use_timeout_fn)


IongIer (commented):
That's all actually. The rest looks good, great choice on what to improve next!

apis/src/app.rs
```diff
@@ -69,29 +69,25 @@ pub fn App() -> impl IntoView {
                     <Route
                         path="/@/:username"
                         view=|| {
-                            view! {
```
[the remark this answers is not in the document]
PenguinWithATie: You are correct, just now realizing how the router works, thanks fixed

apis/src/pages/profile_view.rs
```diff
@@ -1,110 +1,261 @@
...
+}
+
+#[component]
+pub fn DisplayGames(tab_view: ProfileGamesView) -> impl IntoView {
+    let ctx = expect_context::<ProfileGamesContext>();
+    ctx.tab_view.set(tab_view);
+    let username = move || ctx.user.get().map_or(String::new(), |user| user.username);
+    let el = create_node_ref::<html::Div>();
+    let ws = expect_context::<WebsocketContext>();
+    create_effect(move |_| {
```
IongIer, line 224: Same as the other create_effect.

apis/src/providers/sounds.rs
```diff
@@ -54,27 +53,20 @@ impl SoundsSignal {
...
-                },
-                duration,
-            );
             if let Some(buffer) = signal.get_untracked() {
                 let source = context.create_buffer_source().unwrap();
                 source.set_buffer(Some(&buffer));
                 source.set_loop(false);
-                source.start_with_when(offset).unwrap();
+                source
+                    .start_with_when_and_grain_offset_and_grain_duration(0.0, offset, duration)
```
IongIer, line 65: Great change, thank you for getting rid of the the timeout_fn!

apis/src/websockets/api/games_search.rs
```diff
@@ -0,0 +1,51 @@
...
+            pool: pool.clone(),
+        })
+    }
+
+    pub async fn handle(&self) -> Result<Vec<InternalServerMessage>> {
+        let options = self.options.clone();
+        let mut conn = get_conn(&self.pool).await?;
+        let games = Game::get_rows_from_options(&options, &mut conn).await?;
+        let mut game_responses = Vec::new();
+        for game in games.iter() {
```
IongIer, line 31: is the iter() needed here?

IongIer (commented):
This is what I noticed so far, haven't played around with it yet. Will comment some more after I have.

I'm sure that felix will have more stuff to say especially about the backend parts.

apis/src/pages/profile_view.rs
```diff
@@ -1,110 +1,261 @@
...
-    let first_batch_finished = Resource::new(username, move |_| {
-        get_finished_games_in_batches(
-            username(),
-            last_timestamp.get_untracked(),
-            last_id.get_untracked(),
-            5,
-        )
+    let api = ApiRequests::new();
+    let ws = expect_context::<WebsocketContext>();
+    create_effect(move |_| {
```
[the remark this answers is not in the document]
PenguinWithATie: I agree with this, though I am trying to use as intended by the docs so I hope its fine. I couldn't figure out a nicer way to do an api call on page load and also saw this pattern on the tournament page. 

In any case Im hoping passing the signals needed (not related to the websocket) as untracked should minimize any unwanted side effects (if any)

apis/src/pages/profile_view.rs
PenguinWithATie: I will resolve this so I wont get lost with 2 discussion threads

db/src/models/game.rs
```diff
@@ -1081,36 +1068,142 @@ impl Game {
...
-            .select(games::all_columns)
-            .get_results(conn)
-            .await?)
+            ));
+        }
+        if let Some(batch_size) = options.batch_size {
+            query = query.limit(batch_size as i64);
+        }
+        Ok(query.select(games::all_columns).get_results(conn).await?)
     }
```
IongIer, line 1208: Would be good if we could simplify / break this up into smaller functions although it might be complicated to do due to diesel types...Tbh I'm not sure I'd be able to do it myself.

If not simplify at least mark it with a  // TODO: ...

IongIer (approved):
This works well, couldn't find any problems with it other then a few frontend changes to the user profile but I'll do those once we merge it.

If you could refactor the complex function it would be great otherwise we can merge it as is.

IongIer (approved):

PR #319, opened 2024-08-07
add history controls to chat, add analysis button to history

PenguinWithATie:
- add analysis button to history
- add history controls to chat
- moving from history to chat (also chat to history) now maintains history turn, moving to game goes to the last.
- Update deps

Note: removed the mutable reference in `GameStateSignal::show_history_turn` since its needed to use inside the callback and also its not required to update the signal, might be worth changing other functions of this type in a similar way, to make life simpler

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1189,6 +1189,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "displaydoc"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97369cbbc041bc366949bc74d34658d6cda5621039731c6310521892a3a20ae0"
+dependencies = [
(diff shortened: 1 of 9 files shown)


klautcomputing (approved):
Works! We can fix the nitpick or just merge it as is. I will happily deploy once you have merged it.

klautcomputing (approved):

IongIer (changes requested):
This is severely broken. There's a bunch of runtime errors if you try to open chat in the mobile view. + a lot of yellow warnings in the browser console. Not sure I will have time today to investigate where the problem really comes from.....I just checked main, we have some of the errors there as well, so the problem is not completely here but the changes in this stopped the mobile chat from working at least.

IongIer (approved):

PR #325, opened 2024-09-03
Docker dev

kd2718:
Adds support for development with docker compose. The docker script will install all requirements and run all db migrations. This should help speed up development and reduce barrier of entry when developing.


kd2718:
I did not create the wait-for-it.sh script. I grabbed it from here: https://github.com/vishnubob/wait-for-it

IongIer (approved):
Works on my machine!

PR #326, opened 2024-09-04
Fixes multiple tab profile display bug

IongIer:
Fixes the bug, where opening multiple tabs of different player profiles would lead to all of them displaying the games and stats of the last one opened.

However, the fix was done in a quick way and does not follow the same pattern as the other things that get updated by the navigation controller signal.

The reason for this is in the base layout even though you have access to the location url of the page [use_params](https://docs.rs/leptos_router/latest/leptos_router/fn.use_params.html) does not seem to work correctly so I ended up updating the navigation controller further down in the profile component.

We could use this same use_params for the game and tournament, this would get rid of the need for regexes in the base layout however it would need a separate effect further down in the relevant component. 

Is that the better way to go? 
Or should I move the player profile update to base_layout like the rest (this would need another regex for the profile url I think)?
Keep it as is and refactor later?


klautcomputing (approved):
lgtm, tried it and it works

PR #327, opened 2024-09-06
Enable undoing first bug in analysis mode

lakes26:
It is currently impossible to undo the first bug placement on the analysis board. This is desirable for users to explore openings in somewhat of a sandbox like environment. This change enables that option.


IongIer (approved):
Thank you for fixing this!

PR #331, opened 2024-09-17
fix tournament standings display

PenguinWithATie:
- Simplify the code for tournament standings, this also fixes a bug that displays standings incorrectly when ties are present. All unit tests are passing.

- Head2head depends on its position on the tie breaker ordering, calculate purely online now and avoid doing so from player standings

- Sort tournaments by latest updated

### DIFF
--- a/shared_types/src/standings.rs
+++ b/shared_types/src/standings.rs
@@ -43,7 +43,7 @@ impl Standings {
             players: HashSet::new(),
             players_scores: HashMap::new(),
             pairings: HashMap::new(),
-            tiebreakers: vec![],
+            tiebreakers: vec![Tiebreaker::RawPoints],
             players_standings: Vec::new(),
         }
     }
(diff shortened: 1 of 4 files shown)


shared_types/src/standings.rs
```diff
@@ -101,36 +101,24 @@ impl Standings {
...
-            if players.len() > 1 {
-                self.players_standings.remove(i);
-                let mut scores: Vec<(Uuid, f32)> = players
-                    .iter()
-                    .map(|player| {
+    let mut scores: Vec<(Uuid, f32, f32)> = self.players.clone().into_iter().map(|player| {
                         (
-                            *player,
+                            player,
+                            self.get_raw_points(player),
```
IongIer, line 107: Doesn't this calculate the raw_points on the spot? They should already be calculated here.
Couldn't we just do this?
```rs
*self.players_scores.get(&player).unwrap().get(&Tiebreaker::RawPoints).unwrap_or(&0.0)
```

shared_types/src/standings.rs
```diff
@@ -43,7 +43,7 @@ impl Standings {
             players: HashSet::new(),
             players_scores: HashMap::new(),
             pairings: HashMap::new(),
-            tiebreakers: vec![],
+            tiebreakers: vec![Tiebreaker::RawPoints],
```
IongIer, line 46: I think this might lead to problems, as the fronted sends a vec of tiebreakers and when the torunament is created that vec will get added on to this one and we end up with RawPoints in there twice.

It probably doesn't make sense to have a tournament where the first tiebreaker isn't points but this would also prevent that.

shared_types/src/standings.rs
PenguinWithATie: This works to fix the issue, but fails the unit tests, I am fine to change this

IongIer (approved):

PR #333, opened 2024-09-23
simple tournament search by name

PenguinWithATie:
Simple active search by name for tournaments

### DIFF
--- a/apis/src/pages/tournaments.rs
+++ b/apis/src/pages/tournaments.rs
@@ -17,17 +17,34 @@ pub fn Tournaments() -> impl IntoView {
             api.tournament(TournamentAction::GetAll(Abstract));
         };
     });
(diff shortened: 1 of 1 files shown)


apis/src/pages/tournaments.rs
```diff
@@ -17,17 +17,34 @@ pub fn Tournaments() -> impl IntoView {
             api.tournament(TournamentAction::GetAll(Abstract));
         };
     });
+    let search = create_rw_signal(String::new());
```
IongIer, line 20: Just a fyi, create_..signals are getting deprecated in leptos 0.7, better to get used to using RwSignal::new and other Type::new .

IongIer (approved):

IongIer (approved):

PR #345, opened 2024-10-17
Timer refactor/simplify

PenguinWithATie:
Simplify how the timer works:
- eliminate the LowTimeWarning struct
- Avoid the use of memo and create_effect, instead rely on watch and signals (more standard pattern)

Behaviour is identical

### DIFF
--- a/apis/src/components/molecules/live_timer.rs
+++ b/apis/src/components/molecules/live_timer.rs
@@ -6,7 +6,10 @@ use hive_lib::{Color, GameStatus};
 use lazy_static::lazy_static;
 use leptos::*;
 use leptos_router::RouterContext;
-use leptos_use::{use_interval_fn_with_options, utils::Pausable, UseIntervalFnOptions};
+use leptos_use::{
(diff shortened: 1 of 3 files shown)


apis/src/components/molecules/live_timer.rs
```diff
@@ -139,34 +111,40 @@ pub fn LiveTimer(side: Signal<Color>) -> impl IntoView {
...
+    #[allow(unused)]
+    whenever_with_options(
+        user_needs_warning,
+        move |_, _, issued| {
+            let issued = issued.unwrap_or_default();
+            if issued {
+                !should_refresh_warning()
+            } else {
+                sounds.play_sound(SoundType::LowTime);
+                true
```
IongIer, line 127: Why does this closure need to return something? can't the condition here be modified, to me it seems like it doesn't need two branches

PenguinWithATie: the return from the closure is the "issued" flag now instead of getting it from the signal

IongIer (approved):

PR #347, opened 2024-10-22
consolidate cookies

PenguinWithATie:
Refactor and simplify user config

- Dont use server actions (and delete now superfluous types
- Leverage use_cookie api from leptos-use for simpler code
-  All configs now under a single cookie (user_config)
- Gathers values from previous cookies to maintain saved presets


apis/src/providers/config.rs
```diff
@@ -0,0 +1,85 @@
+use std::collections::HashMap;
+use std::str::FromStr;
+
```
klautcomputing, line 3: Remove the 2 empty lines please

apis/src/providers/config.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Config {
+    pub fn get_cookie() -> (Signal<Option<ConfigOpts>>, WriteSignal<Option<ConfigOpts>>) {
+        let exp = Utc::now()
+            .checked_add_signed(chrono::Duration::days(365))
+            .unwrap()
+            .timestamp();
+        use_cookie_with_options::<ConfigOpts, Base64<MsgpackSerdeCodec>>(
+            USER_CONFIG_COOKIE,
+            UseCookieOptions::default()
+                .same_site(SameSite::Lax)
```
klautcomputing, line 35: Why are we doing `Lax` here? Do we need the cookie to be cross-site?

apis/src/providers/config.rs
```diff
@@ -0,0 +1,85 @@
...
+                .secure(true)
+                .expires(exp),
+        )
+    }
+}
+
+impl ConfigOpts {
+    //Could be removed eventually, only here to migrate old cookies
```
klautcomputing, line 43: Do you want to do that in this PR as well?

klautcomputing (changes requested):
That makes the code a lot easier :)

apis/src/providers/config.rs
PenguinWithATie: true, strict will suffice

apis/src/providers/config.rs
PenguinWithATie: I think its better to do that later, for now lets pick up the old cookies (if any) to create the new, so we can maintain saved presets. Though if thats not a big concern, im ok to remove

IongIer (approved):

klautcomputing (approved):

PR #348, opened 2024-10-23
websocket refactor

PenguinWithATie:
- move all api handlers to either `websocket::server_handlers` or `websocket::client_handlers` for clarity/ease of use
- reduce the number of public exports of the websocket module
- pings is now inside lag_tracking (as a private submodule)
- lag tracking is inside websocket  (as a private submodule)
- properly gate to ssr the corresponding modules


apis/src/websocket/client_handlers/response_handler.rs
```diff
@@ -5,14 +5,13 @@ use super::{
...
+use crate::common::{ServerMessage::*, ServerResult, WebsocketMessage};
 use leptos::logging::log;
-use leptos::*;
 
-pub fn handle_response(m: &CommonMessage) {
-    batch(move || match m {
-        CommonMessage::Server(result) => match result {
-            ServerResult::Ok(message) => match *message.clone() {
+pub fn handle_response(m: WebsocketMessage) {
+    match m {
```
IongIer, line 12: The batch here was so that when one of the handle_...updates multiple signals it doesn't trigger multiple updates and it just updates once for all of them.

PenguinWithATie: Ive put it back. To note on leptos 0.7 everything is batched and the api is gone, so they will need to be removed eventually

IongIer (approved):

PR #358, opened 2024-11-04
Adds romanian locales

IongIer:


klautcomputing (approved):
Oh, yes I totally understood everything!

PR #362, opened 2024-11-05
add history timer display

PenguinWithATie:
Display time remaining in timer for all players after each move, when game has ended

cargo format also wanted to format a few unrelated files.


apis/src/components/atoms/history_button.rs
```diff
@@ -73,4 +74,48 @@ fn send_action(action: &HistoryNavigation) {
...
+}
+
+pub fn set_timer_from_response() {
+    let game_state_signal = expect_context::<GameStateSignal>();
+    let timer = expect_context::<TimerSignal>();
+    let is_finished = game_state_signal.is_finished();
+    //don't update timer if game isnt finished or for correspondence games
+    if !is_finished()
+        || matches!(
+            game_state_signal.signal.get().get_game_speed(),
```
IongIer, line 87: game_state_signal.signal.get().game_response.as_ref().is_some_and(|gr| gr.time_mode == TimeMode::RealTime) also works, to me that seems better instead of the match. I might be wrong tho.

apis/src/components/atoms/history_button.rs
```diff
@@ -73,4 +74,48 @@ fn send_action(action: &HistoryNavigation) {
...
+            None | Some(GameSpeed::Correspondence) | Some(GameSpeed::Untimed)
+        )
+    {
+        return;
+    }
+    let turn = game_state_signal
```
IongIer, line 93: It's probably worth getting the signal just once and then using it in all the places it's currently got.

IongIer (approved):

PR #367, opened 2024-11-16
PWA

frisoft:
Make the website installable as a PWA.


Cargo.toml
```diff
@@ -61,6 +61,7 @@ itertools = "0.13.0"
 leptix_primitives = { version = "0.2" }
 tree-ds = {version = "0.1.5", features = ["serde", "compact_serde"] }
 bimap = {version = "0.6.3", features = ["serde"] }
+walkdir = "2.5"
```
frisoft, line 64: I've added this to use directly, but it was already part of the dependencies in Cargo.lock.
Can't do without it, if adding a dependency is not desirable.

apis/assets/site.webmanifest
```diff
@@ -1,14 +1,21 @@
 {
-    "name": "",
-    "short_name": "",
+    "name": "HiveGame",
```
frisoft, line 2: Not completely sure about the name of the Webapp. Should it be just "Hive" or "Hive Game" or something else?

apis/src/pwa_cache.rs
```diff
@@ -0,0 +1,27 @@
+#[cfg(feature = "ssr")]
+use actix_web::{get, HttpResponse, Responder, web};
+use walkdir::WalkDir;
+
+#[cfg(feature = "ssr")]
+#[get("/pwa-cache")]
+pub async fn pwa_cache(site_root: web::Data<String>) -> impl Responder {
```
frisoft, line 7: This might deserve some memoization. But, for now, it should be fine as the tree of assets is quite small and that would be a too-early optimisation.
WDYT?

PenguinWithATie, line 7: the right place for this file is probably apis/src/functions

apis/assets/site.webmanifest
frisoft: I'll keep that name, then. Thank you

apis/src/pwa_cache.rs
frisoft: I moved it under apis/src/functions/. Could you please have a look, @PenguinWithATie?

apis/src/functions/pwa.rs
```diff
@@ -0,0 +1,26 @@
+#[cfg(feature = "ssr")]
+use actix_web::{get, HttpResponse, Responder, web};
+use walkdir::WalkDir;
+
+#[cfg(feature = "ssr")]
+#[get("/pwa-cache")]
+pub async fn cache(site_root: web::Data<String>) -> impl Responder {
```
klautcomputing, line 7: @frisoft You could use https://docs.rs/lazy_static/latest/lazy_static/ so that we only need to compute the cache once. But I am happy to roll it out as is and then when it's getting too slow we can intervene 👍

apis/src/pwa_cache.rs
PenguinWithATie: looks good ty

PenguinWithATie (approved):

apis/src/functions/pwa.rs
frisoft: I'd merge this one and create another PR to introduce the lazy_static.OK?

PR #383, opened 2024-11-22
Adds tournament finish

klautcomputing:


db/src/models/game.rs
```diff
@@ -1197,10 +1197,24 @@ impl Game {
         if let Some(tid) = self.tournament_id {
             let tournament = Tournament::find(tid, conn).await?;
             tournament.ensure_user_is_organizer(user_id, conn).await?;
+            if *new_result == TournamentGameResult::Draw
```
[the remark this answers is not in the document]
klautcomputing: Because TGR also gets set when a game ends by two people playing a game, which in turn can obviously end in a draw :)

apis/src/components/molecules/unplayed_game_row.rs
```diff
@@ -0,0 +1,187 @@
...
+            "Not yet scheduled".to_owned()
+        }
+    };
+    let show_adjudicate_menu = RwSignal::new(false);
+    let toggle_adjudicate = move |_| {
+        show_adjudicate_menu.update(|b| *b = !*b);
+    };
+    let white = move |_| {
+        if user_is_organizer() {
+            let action = TournamentAction::AdjudicateResult(
```
klautcomputing, line 49: @IongIer do we want to make them one function and just pass the result?

apis/src/components/organisms/standings.rs
```diff
@@ -1,67 +1,57 @@
...
-                        players_at_position.iter().map(|(uuid, _, _)| *uuid).collect::<Vec<Uuid>>()
-                    }
-
-                    let:players_at_position
+                    each=move || tournament().standings.results().into_iter().flatten()
+                    key=|(uuid, position, finished, hash)| (
+                        *uuid,
+                        position.clone(),
+                        *finished,
+                        hash.values().map(|x| x.ceil() as i64).sum::<i64>(),
```
IongIer, line 36: This was a sort of hacky solution to fix the reactivity but it seems to work correctly. the values in the map are f32 but I couldn't sum over f32 because f32 doesn't implement hash and the key= requires hahsable elements. When Score row gets called it uses the initial values in the hash

apis/src/components/molecules/tournament_finished_notification.rs
```diff
@@ -0,0 +1,41 @@
...
+        notifications.tournament_finished.update(|t| {
+            t.remove(&tournament().tournament_id);
+        });
+    };
+
+    view! {
+        <div class="flex items-center text-center cursor-pointer dark:odd:bg-header-twilight dark:even:bg-reserve-twilight odd:bg-odd-light even:bg-even-light max-w-fit">
+            <div class="flex relative">
+                <div class=div_class>
+                    <div>{tournament().name} " finished"</div>
```
PenguinWithATie, line 23: Probably another line of text saying "Tournament notification" would be nicer, just seeing "X finished might be a bit confusing

apis/src/components/organisms/standings.rs
PenguinWithATie: I dont get the use of celi here, why not just `hash.values().sum::<f32>() as i64` ?

PenguinWithATie (commented):
EDIT: ion explained to me im stupid, deleted all

IongIer (approved):

PenguinWithATie (approved):

IongIer (approved):

PR #385, opened 2024-11-23
Makes disconnect obvious

IongIer:


PenguinWithATie:
It would be nice to make this even more obvious by adding a [Backdrop Blur](https://tailwindcss.com/docs/backdrop-blur)

IongIer:
> It would be nice to make this even more obvious by adding a [Backdrop Blur](https://tailwindcss.com/docs/backdrop-blur)

It would make it more obvious but it would also make it harder to think about the position while it's reconnecting. Since time is running while you are reconnecting it's better if you can still think about the position without too much distraction.

PenguinWithATie (approved):

PR #386, opened 2024-11-24
Simplfy sounds loading and reactivity

PenguinWithATie:
- Simplifies the code for sounds and also avoids a warning in the console
- Fix a few warnings in pwa.rs


IongIer (approved):
Nice, thank you for throwing away the garbage :)
Seems to work well!

Issue #375, opened 2024-11-20
Make URL clickable in Tournament Description box


FrascoSal:
Currently when a tournament is created, in the description box URL cannot be clicked directly. It should be easy to click and a new tab in the browser with the content should open. This is useful for links to rules of the tournaments or discord, etc.

Example: https://hivegame.com/tournament/9XEdriL-4E-

It would be great if there was a tool similar to the one I see here when I create a new issue in github: (option to have bold, italic, add URL, bullet points, etc). Not essential but would be nice.

PR #389, opened 2024-11-24
add markdown rendering to tournaments

PenguinWithATie:
- Solves https://github.com/hiveboardgame/hive/issues/375 by adding markdown rendering to tournament descriptions. This does not make links clicable by default but allows the user to create them using the [Common Mark](https://commonmark.org/) spec (which has even more capabilities than just links)

- Adds a small edit/preview button that allows the user to preview the markdown they are inputing
- Bumps a few crates to latest


apis/src/pages/tournament_create.rs
```diff
@@ -249,17 +251,35 @@ pub fn TournamentCreate() -> impl IntoView {
...
+                                {move || if is_not_preview_desc() { "Preview" } else { "Edit" }}
+                            </button>
+                            <span class="px-1 font-bold">Description:</span>
+
+                        </div>
+                        <Show
+                            when=is_not_preview_desc
+                            fallback=move || {
+                                view! {
+                                    <div class="prose dark:prose-invert" inner_html=markdown_desc />
```
IongIer, line 269: This needs some extra classes otherwise when you switch to preview the generated html can flow out of it's parent div and push stuff around in weird ways. Something like this seems to work well:

```html
<div class="p-4 w-full break-words prose dark:prose-invert" inner_html=markdown_desc />
```

apis/src/pages/tournament_create.rs
```diff
@@ -249,17 +251,35 @@ pub fn TournamentCreate() -> impl IntoView {
...
+                                }
+                            }
+                        >
+                            <textarea
+                                class="px-3 py-2 w-10/12 leading-tight rounded border shadow appearance-none focus:outline-none"
+                                name="Tournament description"
+                                type="text"
+                                prop:value=tournament.description
+                                placeholder="At least a 50 character description (markdown supported)"
```
IongIer, line 278: Maybe the placeholder should explain how to do markdown links?

apis/src/pages/tournament.rs
```diff
@@ -253,13 +253,11 @@ fn LoadedTournament(tournament: Signal<TournamentResponse>) -> impl IntoView {
...
     view! {
-        <div class="flex flex-col justify-center p-2 w-full">
+        <div class="flex flex-col items-center p-2 w-full">
             <h1 class="w-full max-w-full text-3xl font-bold text-center whitespace-normal break-words">
                 {move || tournament().name}
             </h1>
-            <div class="w-full text-center whitespace-normal break-words">
-                {move || tournament().description}
-            </div>
+            <div class="flex flex-col py-2 prose dark:prose-invert" inner_html=markdown_desc />
```
IongIer, line 260: To stop single line long descriptions from breaking the layout of the page the classes need to change a bit, something like this works, as also noted for the create page
```html
            <div class="p-4 w-full break-words prose dark:prose-invert" inner_html=markdown_desc />
```

apis/src/pages/tournament_create.rs
PenguinWithATie: Added some info to the placeholder and also a link to a common mark cheatsheet

PR #392, opened 2024-11-29
Changes date in pgn file name

klautcomputing:


apis/src/components/atoms/download_pgn.rs
```diff
@@ -85,6 +85,7 @@ fn blob_and_filename(game: StoredValue<GameResponse>) -> (Blob, String) {
         file.push(format!("\n{game_result}\n"));
     }
     let file = file.into_iter().map(JsValue::from).collect::<Array>();
+    let date = game.created_at.format("%+").to_string();
```
IongIer, line 88: Took me a while to figure out why line 56 let date ... is still in there.
Looks good to me :)

klautcomputing: I thought it looked better in the file with the month written out but for the filename I wanted to keep it shorter

PR #396, opened 2024-12-07
fix pwa cookie

PenguinWithATie:
Having the cookie's SameSite attribute as strict doesn't work for the mobile PWA and the cookie gets reset when closing the app

Setting the SameSite as Lax, properly persists the cookie


IongIer (approved):

klautcomputing (approved):
let's try it 👍

PR #410, opened 2025-01-10
Fixes non existing move times breaking history

klautcomputing:
https://doc.rust-lang.org/std/vec/struct.Vec.html#indexing 🙂


IongIer (approved):
Sheep pit

PR #418, opened 2025-01-27
Adds uniqueness to nanoids

klautcomputing:
### Why?
```sql
hive-local=> SELECT nanoid, COUNT(*)
FROM games
GROUP BY nanoid
HAVING COUNT(*) > 1;
   nanoid   | count
------------+-------
 qtR7zXu9QZ |     2
 ```
### Solution
Add unique idx to everything that has nanoids

### DIFF
--- a/db/src/models/challenge.rs
+++ b/db/src/models/challenge.rs
@@ -3,12 +3,13 @@ use crate::{
     models::user::User,
     schema::{
         challenges::{self, nanoid as nanoid_field, opponent_id as opponent_id_field},
+        games::{self, nanoid as game_nanoid},
         users,
     },
     DbConn,
 };
 use chrono::prelude::*;
-use diesel::prelude::*;
+use diesel::{dsl::exists, prelude::*, select};
 use diesel_async::RunQueryDsl;
 use nanoid::nanoid;
 use serde::Serialize;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by IongIer — APPROVED** [review]

File: db/src/models/challenge.rs [hunk 1]
```diff
@@ -85,8 +89,21 @@ impl NewChallenge {
                 error: String::new(),
             });
         };
+        let mut nanoid = String::new();
+        loop {
```
kd2718, line 93: This is good, but we might want to move this into the DB as a trigger in the future. Then the DB can be in charge of the ids and doesn't have to send results back and forth to check for a unique id.

db/src/models/challenge.rs
```diff
@@ -85,8 +89,21 @@ impl NewChallenge {
...
+            nanoid = nanoid!(12);
+            if games::table
+                .filter(game_nanoid.eq(&nanoid))
+                .count()
```
kd2718, line 97: Can we use `exists()` here instead? that is usually a lot faster than `count()`

kd2718 (commented):
Looks good for the most part. Added a few comments.

db/src/models/challenge.rs
klautcomputing: Yeah that's a good idea 👍 but for not it's only during challenge creation so I decided this is the way quicker version. Changed it to an exist as well, which should be even faster.

kd2718 (approved):

PR #428, opened 2025-02-12
simplify auth context type

PenguinWithATie:
- change the signature of get_account, remove the option type from the result as its redundant. also adjust the related context

- use latest i18n instead of git branch


klautcomputing (approved):
👍 
Wow, they were fast with getting that fix in and then released :) I opened an issue about it today.

PR #431, opened 2025-02-18
Printer

klautcomputing:
@pashneal


engine/src/board.rs
```diff
@@ -61,6 +127,141 @@ impl Board {
...
+        let pieces = std::fs::read_to_string("pieces.svg")?;
+        writeln!(writer, "{pieces}")?;
+        writeln!(
+            writer,
+            "<g transform=\"translate({} {})\">",
+            -1.0 * min_x + space_around,
+            -1.0 * min_y + space_around
+        )?;
+
+        for (position, piece) in positions_piece.iter() {
```
klautcomputing, line 195: This needs to be turned into an impl Iter on board so it can be reused from other places

IongIer (approved):

PR #434, opened 2025-02-20
registration works, login sorta works, still very broken

PenguinWithATie:
make pool, identity and uuid async and use leptos extractors instead of requesting context, this change is also probably required in a few other places

add the action field to auth context, simplifies things a lot

make ApiRequests an actual provided context

challenges work now


klautcomputing (approved):
This indeed makes it better!

IongIer (approved):

PR #440, opened 2025-02-27
Adds bots API

klautcomputing:
@nrobinaubertin

First draft of the API, please give me loads of feedback

### DIFF
--- a/apis/src/api/v1/auth/get_identity_handler.rs
+++ b/apis/src/api/v1/auth/get_identity_handler.rs
@@ -1,0 +1,14 @@
+use super::auth::Auth;
+use actix_web::get;
+use actix_web::HttpResponse;
+use serde_json::json;
+
+#[get("/api/v1/auth/id")]
+pub async fn get_identity(Auth(email): Auth) -> HttpResponse {
+    HttpResponse::Ok().json(json!({
+      "success": true,
+      "data": {
+        "user": email,
+      }
+    }))
+}
--- a/apis/src/api/v1/auth/get_token_handler.rs
+++ b/apis/src/api/v1/auth/get_token_handler.rs
@@ -1,0 +1,78 @@
+use super::encode::{jwt_encode, Bot};
+use super::jwt_secret::JwtSecret;
+use actix_web::post;
+use actix_web::web::{Data, Json};
+use actix_web::HttpResponse;
+use argon2::{
+    password_hash::{PasswordHash, PasswordVerifier},
+    Argon2,
+};
+use db_lib::get_conn;
+use db_lib::{models::User, DbPool};
+use serde_json::json;
+
+#[post("/api/v1/auth/token")]
+pub async fn get_token(
+    Json(bot): Json<Bot>,
+    pool: Data<DbPool>,
+    jwt_secret: Data<JwtSecret>,
+) -> HttpResponse {
+    let mut conn = match get_conn(&pool).await {
+        Ok(conn) => conn,
+        Err(e) => {
+            return HttpResponse::InternalServerError().json(json!({
+            "success": false,
+            "data": {
+              "message": "Internal server error",
+            }}))
+        }
+    };
+    let user = match User::find_by_email(&bot.email, &mut conn).await {
+        Ok(user) => user,
+        Err(e) => {
+            return HttpResponse::NotFound().json(json!({
+              "success": false,
+              "data": {
+                "message": "No user found by that email address",
+              }
+            }))
+        }
+    };
+    let argon2 = Argon2::default();
+    let stored_pw = match PasswordHash::new(&user.password) {
+        Ok(pw) => pw,
+        Err(e) => {
+            return HttpResponse::InternalServerError().json(json!({
+            "success": false,
+            "data": {
+              "message": "Internal server error",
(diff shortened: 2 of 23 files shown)


apis/src/api/v1/bot/play.rs
```diff
@@ -0,0 +1,64 @@
...
+use hive_lib::State;
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use shared_types::GameId;
+
+#[derive(Serialize, Deserialize)]
+struct PlayRequest {
+    game_id: GameId,
+    piece: String,
+    position: String,
```
nrobinaubertin, line 22: I suppose that an example of piece/position would be "wP"/"bL-". It would be nicer to expect a standard move ("wP bL-") and cut it yourself

apis/src/api/v1/auth/get_identity_handler.rs
```diff
@@ -0,0 +1,14 @@
+use super::auth::Auth;
+use actix_web::get;
+use actix_web::HttpResponse;
+use serde_json::json;
+
+#[get("/api/v1/auth/id")]
```
nrobinaubertin, line 6: Why do we need that endpoint ?

apis/src/api/v1/auth/get_token_handler.rs
```diff
@@ -0,0 +1,78 @@
...
+use actix_web::HttpResponse;
+use argon2::{
+    password_hash::{PasswordHash, PasswordVerifier},
+    Argon2,
+};
+use db_lib::get_conn;
+use db_lib::{models::User, DbPool};
+use serde_json::json;
+
+#[post("/api/v1/auth/token")]
```
nrobinaubertin, line 14: So everyone can use the API to play as a bot ? (I don't see any filtering of bot user accounts). But I understand that this PR is a stepping stone.

apis/src/api/v1/bot/games.rs
```diff
@@ -0,0 +1,63 @@
...
+    Json(req): Json<GameRequest>,
+    Auth(email): Auth,
+    pool: Data<DbPool>,
+) -> HttpResponse {
+    match get_games(&email, req.selector, pool).await {
+        Ok(games) => HttpResponse::Ok().json(json!({
+          "success": true,
+          "data": {
+            "user": email,
+            "games": games,
```
nrobinaubertin, line 41: So, if I understand correctly, the whole game object is going to be serialized ? If I look at `models::Game` I see that we have the players UUIDs but not their name. Shouldn't we give that information ?

nrobinaubertin (commented):
Only high-level comments since I'm no rust dev.  
We should write/generate an API doc in the future

apis/src/api/v1/auth/get_identity_handler.rs
klautcomputing: This is basically a "health check" for you token. Helped me debug things, might help others. We might also use this to return some meaningful things about your account, like elo, games played, ...

apis/src/api/v1/bot/play.rs
klautcomputing: Sure, happy to make that change.

apis/src/api/v1/auth/get_token_handler.rs
klautcomputing: @PenguinWithATie suggested to add a column to `User`. I think that's what I am going to do before shipping it.
We also need the `bot` field to advertise bot challenges and to indicate that an account is a bot.

apis/src/api/v1/bot/games.rs
klautcomputing: Currently yes, but that's just because I was a) lazy b) didn't know which fields we would like to keep/add. Once we have that I am going to write a serializer that just send all the necessary info. But be careful: usernames can change. So using UUID might be the most reliable thing. I can write you a small uuid -> user (profile) api endpoint.

nrobinaubertin: And endpoint to translate uuid to user would be interesting. Especially if you give more information about the user that way: like their elo rating

apis/src/api/v1/auth/get_token_handler.rs
```diff
@@ -0,0 +1,57 @@
...
+pub async fn get_token(
```
frisoft, line 16: Wasn't simper to manage the Bot authentication via an API_KEY stored in the bot account (user) record?
Authentication on both sides would have been much simpler.

Anyway, this works as well.

frisoft (approved):
There are some conflicts.

apis/src/api/v1/bot/challenges.rs
```diff
@@ -0,0 +1,132 @@
...
+        Err(e) => HttpResponse::Ok().json(json!({
+          "success": false,
+          "data": {
+            "error": e.to_string(),
+          }
+        })),
+    }
+}
+
+#[get("/api/v1/bot/challenge/accept/{nanoid}")]
```
frisoft, line 45: This is an endpoint to accept a challenge. Isn't it?
Shouldn't it be a POST instead of a GET?

IongIer (approved):

PR #455, opened 2025-03-20
Admin banner

PenguinWithATie:


klautcomputing:
@PenguinWithATie I think just cherry picking the files/changes you need from a clean branch might be quicker here 😄

PenguinWithATie:
@klautcomputing @IongIer  lets get this one in also

apis/src/components/atoms/create_challenge_button.rs
```diff
@@ -10,7 +10,10 @@ pub fn CreateChallengeButton(
 ) -> impl IntoView {
     let config = expect_context::<Config>().0;
     let icon_data = move |color_choice: ColorChoice| match color_choice {
-        ColorChoice::Random => (icondata::BsHexagonHalf, "w-full h-full"),
+        ColorChoice::Random => (
+            icondata::BsHexagonHalf,
+            "w-full h-full stroke-1 stroke-black",
```
IongIer, line 15: the two extra styles need to go actually just w- and h- need to remain here

apis/src/components/atoms/piece.rs
```diff
@@ -1,6 +1,7 @@
 use crate::common::{MoveConfirm, TileDesign, TileDots, TileRotation};
 use crate::common::{PieceType, SvgPos};
 use crate::components::organisms::analysis::AnalysisSignal;
+//use crate::components::organisms::analysis::AnalysisSignal;
```
IongIer, line 4: this can go as well

apis/src/components/molecules/rl_banner.rs
```diff
@@ -1,34 +1,14 @@
 use leptos::prelude::*;
 
 #[component]
-pub fn RlBanner() -> impl IntoView {
+pub fn RlBanner(title: String, content: String) -> impl IntoView {
+    //Unrestricted markdown, allows images
+
+    let markdown_desc = move || markdown::to_html(&content);
```
IongIer, line 7: We might want something like:

```
    let markdown_desc = move || {
        markdown::to_html_with_options(
            &content,
            &Options {
                compile: CompileOptions {
                    allow_dangerous_html: true,
                    allow_dangerous_protocol: true,
                    ..CompileOptions::default()
                },
                ..Options::default()
            },
        )
        .unwrap()
    };
    ```
    As long as we trust our admins this should be fine. That way it allows us to have links that look like links and open in new tabs by including actual html in there. there is no target="_blank" for markdown links for example.

File: apis/src/pages/admin.rs [hunk 4]
```diff
@@ -1,17 +1,119 @@
...
+                                            class="px-3 py-2 w-10/12 leading-tight rounded border shadow appearance-none focus:outline-none"
+                                            name="title"
+                                            type="text"
+                                            prop:value=title
+                                            placeholder="banner title"
+                                            on:input=update_from_input(title)
+                                        />
+                                    </div>
+
+                                    <textarea
```
IongIer, line 77: This doesn't change color along when switching light/dark. We could do another tailwind apply for textarea as well.

IongIer (approved):

PR #475, opened 2025-04-12
Switches around default and fallback on play page

IongIer:


klautcomputing (approved):
🤯 what's going on here? I can't believe this works...

PR #483, opened 2025-04-22
Fixes notification issues

IongIer:
- uses Transition in notification dropdown
- started/finished notifications only get sent to users in the tournament
- fixes bug where wrong notification is send on finished tournament

TODO for future PR: don't fetch all tournaments from the db each time a notification comes in


klautcomputing (approved):
:shipit:

PR #484, opened 2025-04-22
Allows users to log in with username as well

klautcomputing:


apis/src/pages/login.rs
```diff
@@ -32,8 +32,7 @@ pub fn Login(#[prop(optional)] extend_tw_classes: &'static str) -> impl IntoView
                         class="px-3 py-2 w-full leading-tight rounded border shadow appearance-none focus:outline-none"
                         name="email"
                         id="email"
-                        type="email"
-                        inputmode="email"
+                        type="text"
                         autocomplete="email"
                         placeholder=move || t_string!(i18n, user_config.login.email)
```
IongIer, line 37: At the very least placeholder needs to changed so it communicates with the user that they can use username as well

IongIer (approved):

PR #486, opened 2025-04-23
Transitions away from warnings

IongIer:


klautcomputing (approved):
LGTM but the AI overlord thinks you should:

Considerations Before Merging:

    Testing:
        Ensure that the <Transition> component works as expected in all scenarios where <Suspense> was replaced. Verify that the fallback behavior, animations, or transitions do not introduce regressions.
        Test all affected pages (InviteUser, OnlineUsers, EditBanner, ChallengeView, Home, ProfileView, and Tournaments) to ensure they render without errors.

    Code Consistency:
        If the project uses <Transition> instead of <Suspense> going forward, ensure this is documented for consistency in future development.

    Code Review Feedback:
        If any reviewers have raised concerns or suggestions, ensure these have been addressed before merging.

PR #490, opened 2025-04-24
Adds move query to game page

IongIer:
?move=x loads the game at that move, if it's not parsable  or out of bounds of the game it just loads normally at the last move like without it
https://hivegame.fly.dev/game/3gkRHOu4XYDm?move=55
history buttons update the url to reflect the current move


PenguinWithATie:
it might be worth it to try a [query_signal](https://docs.rs/leptos_router/latest/leptos_router/hooks/fn.query_signal.html) to avoid some of the manual handling of the url state

klautcomputing (approved):

PR #492, opened 2025-05-03
Add hive-hydra

frisoft:
Add hive-hydra as a package in the repo.


klautcomputing (approved):
As a typical German, I am gonna start with the criticism :)

I looked at the code, there are things I would have done differently, if you want examples: 
- I am not a big yaml fan
- We could have retries for failed request, especially when people don't run it locally
- we should maybe add a time out then we run external commands

But  the code is nice and from what I can tell, this seems to work fine (for one bot, haven't tried out multiple yet). So I am really happy! :)

My main ask would be to quiet down the bot. I think to run this in production it's a bit too noisy for my liking, I would like to only be informed about things that go wrong and assume: no news is good news. 

So it would be nice if we could add a another flag in the config which sets the debug level. Feel free to merge away. And we do this in a follow PR. 

And again: Thank you!

frisoft:
> As a typical German, I am gonna start with the criticism :)

As a typical Italian, I will complain about your criticisms. Joking :-)
I appreciate that you reviewed the PR and commented on it.
Thank you

> 
> I looked at the code, there are things I would have done differently, if you want examples:
> 
> * I am not a big yaml fan
Really? I love YAML as it is both readable and powerful with complex data structures like list of structs.
Maybe because, coming from Ruby and Ruby on Rails, I'm used to it.

> * We could have retries for failed request, especially when people don't run it locally
Definitely. If the login fails, it is never retried. I'll fix this in a new PR.

> * we should maybe add a time out then we run external commands
Got what you mean. I'll look into this.

> 
> But the code is nice and from what I can tell, this seems to work fine (for one bot, haven't tried out multiple yet). So I am really happy! :)
> 
> My main ask would be to quiet down the bot. I think to run this in production it's a bit too noisy for my liking, I would like to only be informed about things that go wrong and assume: no news is good news.
> 
> So it would be nice if we could add a another flag in the config which sets the debug level. Feel free to merge away. And we do this in a follow PR.

Sure, I can reduce the noise and add a new configuration option for the debug level.

> 
> And again: Thank you!

I'll merge this and open another one for the improvements.

Thank you

PR #496, opened 2025-05-05
stores for challenge params, put them in a cookie

PenguinWithATie:
ChallengeParams are in a store, and with a cleaner underlying type it allows it to be serialized and put in a cookie (and hence persist them)

Note: the close event of the dialog must fire for the cookie to be updated

### DIFF
--- a/apis/src/components/atoms/input_slider.rs
+++ b/apis/src/components/atoms/input_slider.rs
@@ -26,3 +26,29 @@ pub fn InputSlider(
         />
     }
 }
+
+#[component]
+pub fn InputSliderWithCallback(
+    signal: Signal<i32>,
+    callback: Callback<i32>,
+    name: &'static str,
+    #[prop(into)] min: Signal<i32>,
+    #[prop(into)] max: Signal<i32>,
+    step: i32,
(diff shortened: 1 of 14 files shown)


apis/src/components/atoms/simple_switch.rs
```diff
@@ -26,3 +26,27 @@ pub fn SimpleSwitch(
         </label>
     }
 }
+
+#[component]
+pub fn SimpleSwitchWithCallback(
+    checked: Signal<bool>,
+    #[prop(optional)] disabled: Signal<bool>,
+    action: Callback<()>,
```
IongIer, line 34: SimpleSwitch already had an optional callback. Couldn't that have been used instead? They seem identical.

apis/src/components/atoms/input_slider.rs
```diff
@@ -26,3 +26,29 @@ pub fn InputSlider(
         />
     }
 }
+
+#[component]
+pub fn InputSliderWithCallback(
```
IongIer, line 31: Almost the same feedback as SimpleSwitch, can't this just take an optional callback instead of making an extra component?

PR #512, opened 2025-06-21
Magically fixes runtime error

IongIer:
Don't understand why but this stops the runtime error when searching for players. Won't fix the slowness unfortunately, that might be db related.


klautcomputing (approved):
What?!

Issue #460, opened 2025-03-27
Allow to preselect pieces


KES777:
During the game it is not allowed to preselect pieces

![Image](https://github.com/user-attachments/assets/bc134055-ce18-4539-8e10-bb687f7e6244)

But this is crucial during 1+2 games. 

While opponent is thinking I can prepare my next move.

TODO: Allow to preselect pieces while opponent is thinking.

KES777:
I spoke with Kaur yesterday and he told me that this is configurable via settings. Though it would be better to have it enabled by default.

@IongIer Sorry for complains, I have seen that it was resolved but was not aware that I should enable the configuration for this :D

KES777:
If I understand him correct, he hated it when it was by default but it was not configurable.

But if it will be by default and configurable it allows him to turn preselect feature off (do not use default).
But enabled it by default (players who just registered) gives them opportunity to see functionality exists and disable it if they do not like.

Or, probably, "tour" could be implemented to guide users through "Config" for the first time.

But this is nuances you can ignore and close the ticket. Thank you very much for your work on this.

KES777:
*btw. would it be good to implement "Auto disable move confirmation if user low on time and passes configured value"?

Eg. if he configured 15 seconds and he has 10 second left, then move confirmation is disabled.

PR #515, opened 2025-06-27
Allows players to preselect piece on opponent turn

IongIer:
Implements #460


klautcomputing (approved):
Works well, pretty intuitively and I like the code as well!

PR #531, opened 2025-07-03
Fixes classes not getting applied to icon

IongIer:


klautcomputing (approved):
If it's stupid and it works it's not stupid.

PR #535, opened 2025-07-05
Allow bots to login via username

nrobinaubertin:
I did this change blindly without running any tests (but it seems simple enough)

### DIFF
--- a/apis/src/api/v1/auth/get_token_handler.rs
+++ b/apis/src/api/v1/auth/get_token_handler.rs
@@ -41,7 +41,7 @@ async fn get_token_helper(
 ) -> Result<String> {
     let mut conn = get_conn(&pool).await?;
 
-    let user_result = User::find_by_email(&bot.email, &mut conn).await?
+    let user_result = User::find_by_email(&bot.email, &mut conn).await;
     let user = if let Ok(user) = user_result {
         user
     } else {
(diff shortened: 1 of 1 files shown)


apis/src/api/v1/auth/get_token_handler.rs
```diff
@@ -40,7 +40,14 @@ async fn get_token_helper(
     pool: Data<DbPool>,
 ) -> Result<String> {
     let mut conn = get_conn(&pool).await?;
-    let user = User::find_by_email(&bot.email, &mut conn).await?;
+
+    let user_result = User::find_by_email(&bot.email, &mut conn).await?
```
klautcomputing, line 44: This won't work. If the bot cannot be found via email, because you used the botname the function will automatically bubble up the error returned because of the "?" operator at the end.
Compare how we do it for normal login here: https://github.com/hiveboardgame/hive/blob/main/apis/src/functions/auth/login.rs#L17-L27

klautcomputing (changes requested):

apis/src/api/v1/auth/get_token_handler.rs
nrobinaubertin: So the correct way to do it is to add `.map_err(ServerFnError::new);` after the await ?

klautcomputing: No, you don't want to bubble up the result via the "?" operator. And I think you are missing a ";" there as well. So:

```suggestion
    let user_result = User::find_by_email(&bot.email, &mut conn).await;
```

nrobinaubertin: Okay, I think I get it now. I committed your proposed change

PR #537, opened 2025-07-06
Calculates initial profile batch size based on viewport

### LINKED ISSUE #485: Scroll does not work on finshed games page if I click any button on filter
Image

It would be nice if filter by player will be implemented here

IongIer:
Closes #485 by loading initial batch of games based on current screensize.


klautcomputing (approved):
seems to work 😄

Issue #508, opened 2025-06-13
After 30 min of analysis the page becomes too slow on mobile.


KES777:
When I do analysis on mobile, after ~30 mins the page becomes so slow when I press next/prev buttons.
After opening analysis again all becomes fine.

IongIer:
Has this improved for you or is it the same?

KES777:
It seems it still slow. I fill a lag around 1 second after long review when pressing previous/next turn.

PR #545, opened 2025-07-08
Changes how analysis loads games

IongIer:
1.Can now open analysis of games in a separate tab.

2.Just changing from hivegame.com/game/:nanoid to hivegame.com/analysis/:nanoid is enough to load the game in analysis.  This is prevented for relatime games if you are one of the players, it just loads a blank analysis in that case.

3.Also clones the large analysis tree structures less hopefully this helps with #508


PenguinWithATie (approved):
Nice one, thank you!

klautcomputing (approved):
Works 👍

PR #546, opened 2025-07-08
Returns early if game is already finished

IongIer:
Possible fix for the issue with resigning vs the bot.


klautcomputing (approved):
We could even move it to just after getting the game.

klautcomputing (approved):

PR #562, opened 2025-07-13
Adds bot challenge creation endpoint

### LINKED ISSUE #561: Enable creating open challenges from the bot API
The current bot API allows to list and accept incoming challenges but not their creation. It seems interesting to me to add the possibility for a bot to broadcast it's willingness to play in some time controls by adding the possibility of creating challenges.  

I imagine the the endpoint should be `#[post("/api/v1/bot/challenges/")]`, specifying the time control, if the challenge is open or private, rated or unrated.

IongIer:
Closes #561 

Json with this stuff. 
It could be changed/simplified not sure all of them make sense to have, like the rating bands. Maybe even allowing correspondence is not needed. What do you think?

pub enum BotTimeControl {
    Untimed,
    RealTime { base: u32, increment: u32 }, //min sec
    Correspondence { mode: CorrespondenceMode, days: u32 }, // DaysPerMove or TotalTimeEach
}

pub struct BotChallengeRequest {
    pub game_type: GameType, // Base, MLP... base needs to be unrated
    pub visibility: ChallengeVisibility, //Public, Direct, Private (not sure private makes sense for bots)
    pub opponent: Option<String>, // Direct needs opponent Username
    pub color_choice: ColorChoice, // White, Black, Random
    pub time_control: BotTimeControl,
    pub rated: bool,
    pub band_upper: Option<i32>, //Min max rating
    pub band_lower: Option<i32>,
}

### DIFF
--- a/apis/src/api/v1/bot/challenges.rs
+++ b/apis/src/api/v1/bot/challenges.rs
@@ -1,24 +1,129 @@
 use crate::api::v1::auth::Auth;
-use crate::api::v1::messages::send::send_challenge_messages;
+use crate::api::v1::messages::send::{send_challenge_creation_message, send_challenge_messages};
 use crate::responses::{ChallengeResponse, GameResponse};
-use crate::websocket::busybee::Busybee;
-use crate::websocket::WsServer;
+use crate::websocket::{busybee::Busybee, WsServer};
 use actix::Addr;
 use actix_web::{
-    get,
-    web::{Data, Path},
+    get, post,
+    web::{Data, Json, Path},
     HttpResponse,
 };
 use anyhow::Result;
 use db_lib::{
     get_conn,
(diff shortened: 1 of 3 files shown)


nrobinaubertin:
> Json with this stuff. It could be changed/simplified not sure all of them make sense to have, like the rating bands. Maybe even allowing correspondence is not needed. What do you think?

I agree that correspondence doesn't mean a lot for a bot, but it's maybe easier to keep for consistency on your side ? I won't use it.  
It could mean that someone uses a server to eventually do a lot of compute on a move, once a day. So that way it could create a "correspondence bot". It would be almost the same as untimed though.

The only thing I see against having more options is the maintenance burden on your side.

klautcomputing (approved):
Let's do private challenges going forward

PR #606, opened 2025-08-04
Explain tiebreakers

klautcomputing:


apis/src/components/organisms/standings.rs
```diff
@@ -1,16 +1,90 @@
...
+                *current = if *current == Some(current_tb.clone()) {
+                    None
+                } else {
+                    Some(current_tb)
+                };
+            });
+        });
+    };
+    
+    let explanation = match &tiebreaker {
```
IongIer, line 36: you can't use the t macro's like this. it needs to be in a closure otherwise it's not reactive to when you change the language

apis/src/components/organisms/standings.rs
```diff
@@ -1,16 +1,90 @@
...
+        </th>
+    }
+}
+
 #[component]
 pub fn Standings(tournament: Signal<TournamentResponse>) -> impl IntoView {
     let tiebreakers = tournament.with_untracked(|t| t.tiebreakers.clone());
+    let open_tooltip = RwSignal::new(None::<Tiebreaker>);
+    
+    // Close tooltip when clicking outside
```
IongIer, line 77: look for on_click_outside elsewhere in the codebase and do that. This only closes it if you click somewhere that's part of the standings table

IongIer (commented):
Please show up in pull requests.

PR #620, opened 2025-08-13
Refactors board, updates viewbox via unified request_animation_frame

IongIer:
Refactors board so that all viewbox updates happen per animation frame instead of each being throttled differently and updating the viewbox more than once per animation frame.


klautcomputing (approved):
Code looks good and from playing around with it, it seems to work fine :)

PR #621, opened 2025-08-14
Websockets that dont rely on actix as much (POC)

PenguinWithATie:
- Use the new websocket paradigm made available on leptos 0.8
- Use a tokio watch channel to broadcast messages to all clients
- Client api just needs to have the clonable sender (from futures::mpsc) to send messages

#### Note: no changes to handler or APIs yet, this is mearly a POC to show it works and get some feedback (the DbgMsg on the client request enum should go away)

Pros of moving forward:
- Less dependance on actix
- Use "standard" sender/reciever types which is the norm in rust async
- Since the server notifications rely on a much thinner type(Rather than CAM model), its potentially faster

Cons:
- you tell me


IongIer:
Cons: We will have 3 ways of doing things, old ws new ws and plain server fn's for a while at least. That said the old ws has some problems we need to fix.

PenguinWithATie:
Closing this. Some ideas here might work, but code is old and has issues. Hopefully I (or some one else) will give it a better try down the line

Issue #603, opened 2025-08-02
Allow to filter games by player name


KES777:
<img width="2220" height="1461" alt="Image" src="https://github.com/user-attachments/assets/415fbcdf-cf1f-453e-8dfe-c93c896dde0c" />

It would be nice to find games against specific player.

As another improvement it would be nice to update statistics by counting only not filtered games.

Eg. if inside filter I input `Stepanzo`, then total/win/loses will be updated to reflect only my games with `Stepanzo`.

PR #645, opened 2025-09-14
feat: add opponent username filter in games filter

gfmota:
This PR adds opponent username filter via a string input in profile page's game filter.

#603 

Captura de Tela 2025-09-14 às 09 39 33

### DIFF
--- a/apis/src/components/organisms/games_filter.rs
+++ b/apis/src/components/organisms/games_filter.rs
@@ -4,11 +4,14 @@ use crate::pages::profile_view::tab_from_path;
 use crate::providers::{load_games, FilterState, GamesSearchContext};
 use hive_lib::Color;
 use leptos::{html, prelude::*};
+use leptos::ev::Event;
+use leptos::leptos_dom::helpers::debounce;
 use leptos_i18n::I18nContext;
 use leptos_icons::*;
 use leptos_router::hooks::use_location;
 use leptos_use::{on_click_outside_with_options, OnClickOutsideOptions};
 use shared_types::{GameProgress, GameSpeed, ResultType};
+use std::time::Duration;
 
 #[derive(Clone, Copy)]
 pub enum TriStateType {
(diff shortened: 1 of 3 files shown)


apis/src/providers/games_search_context.rs
```diff
@@ -84,10 +86,11 @@ pub fn load_games(
         color: filters.color,
         result: filters.result,
     });
+    let player2 = filters.opponent.clone().map(|username| PlayerFilter { username, color: None, result: None });
```
gfmota, line 89: Currently, it only works if it is an exact match, for example: "test_user" does not match "test_user2"
What do you guys think?

apis/src/components/organisms/games_filter.rs
```diff
@@ -305,6 +316,20 @@ pub fn GamesFilter(username: String, ctx: GamesSearchContext) -> impl IntoView {
                                     }}
                                 </FilterButton>
                             </div>
+
```
IongIer, line 319: I think it would be good to reuse UserSearch here and use that to select the opponent, that way you only can select valid opponents. You would need to add a new user action or modify UserRow in some other way to make it selectable

apis/src/components/organisms/games_filter.rs
```diff
@@ -169,6 +173,13 @@ pub fn GamesFilter(username: String, ctx: GamesSearchContext) -> impl IntoView {
         };
         current_filters == reset_target
     });
+    let opponent_debounced_search = debounce(Duration::from_millis(100), 
+        move |ev: Event| {
+            opponent.set(event_target_value(&ev));
```
IongIer, line 178: Instead of the effect it's probably better to update pending in here.

gfmota: Good idea, an unnecessary state, thanks

apis/src/components/organisms/games_filter.rs
gfmota: Didn't know about the UserSearch, I'm going to give it a try

PR #648, opened 2025-10-06
remove client side check from timer becoming red on timeout

king-of-the-hive:
As discussed in [here](https://discord.com/channels/1194638121866907769/1421935294663491654/1421935294663491654), there's no necessity in two signals here: by keeping the one coming from server we avoid the confusion when player sees the opponent's timeout coming from client but later sees that server didn't register.

Tested locally a few times, visually seems to work just fine, not sure how to properly test since it's the matter of less than a second.

### DIFF
--- a/apis/src/components/molecules/live_timer.rs
+++ b/apis/src/components/molecules/live_timer.rs
@@ -142,7 +142,7 @@ pub fn LiveTimer(side: Signal<Color>) -> impl IntoView {
             class=move || {
                 format!(
                     "flex resize h-full select-none items-center justify-center text-xl md:text-2xl lg:text-4xl {}",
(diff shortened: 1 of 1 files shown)


IongIer (commented):
Is the signal used for anything else? should be removed if it isn't

IongIer (approved):

king-of-the-hive:
> Is the signal used for anything else? should be removed if it isn't. I checked. Hmm yeah maybe its good to keep the other use of time_is_zero

It's used in one more place. I'm not too sure if this one can be replaced with timed_out() 

https://github.com/hiveboardgame/hive/blob/4f25e6e6c9ff49d4895c2f2801a94a5ef5ffd687/apis/src/components/molecules/live_timer.rs#L116

PR #649, opened 2025-10-08
adds rating graph into stats modal

king-of-the-hive:
Just a very simple graph to show player's progress over time.

Uses `leptos-chartistry` that allows for automatic axis' ticks etc.

image

On mobile there will be a small horizontal scroll for the svg (I hope that's fine, let me know if this is not ok). This is just to have a bit more pixels for the chart, it will be already small for folks with 200+ games.

image

When not enough games:

image

image

Unfortunately I couldn't get the tooltip on hover over working because that's a dialog and the mouse coordinates are messed up. Plus it would be not great for such small container.

In the future I can see how this could be a separate page where you can click different speeds or compare yourself with other players, so just a first step towards that.


apis/src/components/atoms/rating_history.rs
```diff
@@ -0,0 +1,102 @@
...
+                                            )
+                                            left=TickLabels::aligned_floats()
+                                            bottom=TickLabels::timestamps()
+                                            inner=[
+                                                AxisMarker::left_edge().into_inner(),
+                                                AxisMarker::bottom_edge().into_inner(),
+                                                XGridLine::default().into_inner(),
+                                                YGridLine::default().into_inner(),
+                                            ]
+                                            series=Series::new(|data: &RatingPoint| data.datetime)
```
klautcomputing: Here we can just use RatingHistoryEntry.updated_at

db/src/models/rating.rs
```diff
@@ -102,6 +102,27 @@ impl Rating {
             .await?)
     }
 
+    pub async fn history_for_uuid(
+        uuid: &Uuid,
+        game_speed: &GameSpeed,
+        conn: &mut DbConn<'_>,
+    ) -> Result<Vec<Self>, DbError> {
+        let game_speed = match game_speed {
+            GameSpeed::Untimed => GameSpeed::Correspondence.to_string(),
```
klautcomputing: Untimed games are always unranked/unrated so they don't make sense to be grouped into Correspondence in my mind.

db/src/models/rating.rs
```diff
@@ -102,6 +102,27 @@ impl Rating {
...
+            _ => game_speed.to_string(),
+        };
+
+        // println!("Fetching history for UUID: {}, GameSpeed: {}", uuid, game_speed);
```
klautcomputing: Please clean up your code :)

db/src/models/rating.rs
```diff
@@ -102,6 +102,27 @@ impl Rating {
...
+
+        let rt = ratings_table
```
klautcomputing: Why not just return this right away?

klautcomputing (changes requested):
I can't really say anything about the looks because it doesn't work for me. I only get the "not enough games played" text.
I left you some comments when it comes to the code :)

PR #652, opened 2025-10-19
adds rating graph into stats modal

king-of-the-hive:
Better version of the previous attempt to build a very simple graph to show player's progress over time.

(I misunderstood that `ratings` is only for the latest rating. History is in `games`. In this version, rating for the game is calculated as `rating before the game plus the change after the game`)

Uses `leptos-chartistry` that allows for automatic axis' ticks etc. Size of the graph depends on whether the screen is vertical.

I think I addressed all comments from the previous version so please let me know if I missed anything.

Some screenshots.

image

image

(should look better with normal games number and distribution over time)

image

image

image


apis/src/responses/rating_history.rs
```diff
@@ -0,0 +1,51 @@
...
+cfg_if::cfg_if! { if #[cfg(feature = "ssr")] {
+use db_lib::{
+    models::{Game},
+    DbConn,
+};
+use shared_types::GameSpeed;
+use uuid::Uuid;
+use anyhow::Result;
+impl RatingHistoryResponse {
+    pub async fn get_rating_history_from_uuid_and_speed(id: &Uuid, game_speed: &GameSpeed, conn: &mut DbConn<'_>) -> Result<Vec<Self>> {
```
IongIer: Imo something along these lines this is better/nicer without unwrap_or
```rs
impl RatingHistoryResponse {
    pub async fn get_rating_history_from_uuid_and_speed(
        id: &Uuid,
        game_speed: &GameSpeed,
        conn: &mut DbConn<'_>,
    ) -> Result<Vec<Self>> {
        //Need to decide on a name, maybe this? It's not named as in my other suggestion, dunno up to you not good at deciding on names
        let games = Game::get_rated_games_for_player_and_speed_sorted(*id, game_speed, conn).await?;

        let out = games
            .into_iter()
            .filter_map(|g| {
                if g.white_id == *id {
                    Some((g.white_rating?, g.white_rating_change?, g.updated_at))
                } else if g.black_id == *id {
                    Some((g.black_rating?, g.black_rating_change?, g.updated_at))
                } else {
                //can't be the case but compiler doesn't know
                    None
                }
            })
            .map(|(r, dc, ts)| RatingHistoryResponse {
                rating: (r + dc).floor() as u64,
                updated_at: ts,
            })
            .collect();

        Ok(out)
    }
}
```

IongIer (changes requested):
My feedback, pre trying it out. Let's see what @klautcomputing thinks as well

king-of-the-hive:
I swapped the logic to render on the per day basis as discussed on discord. Please try with your data, I don't have enough to test all edge cases. But I think it should be even nicer now.

IongIer (changes requested):

king-of-the-hive:
I moved bucketing to front. I think all should be fine now. Lmk!

king-of-the-hive:
@klautcomputing I moved all bucketing/calculations to the backend, pls check

apis/src/components/atoms/rating_history.rs
```diff
@@ -0,0 +1,171 @@
...
+        Period::Day
+    } else if duration < Duration::days(365) {
+        Period::Month
+    } else {
+        Period::Year
+    };
+    TickLabels::from_generator(Timestamps::<Utc>::from_period(period))
+}
+
+fn get_browser_timezone() -> String {
```
IongIer: One way to avoid having to send the user's local time to the backend thus avoiding the need for js_sys and chrono_tz would be to do the bucketing in the backend based on the servertime and then in the frontend map that to the local time, similar to how we do it in other places where we display a user's local time, like in Calendar. Let's see what @klautcomputing thinks about this as he requested the bucketing in the backend.

apis/src/components/atoms/rating_history.rs
```diff
@@ -0,0 +1,171 @@
...
+                                             stroke: #1a181845;
+                                            }
+                                            ._chartistry_tooltip {
+                                             font-family: 'Inter', system-ui, sans-serif !important;
+                                             font-size: 14px !important;
+                                             color: #111 !important;
+                                             background-color: #fff !important;
+                                             border-radius: 6px !important;
+                                            }
+                                            .dark ._chartistry_tooltip {
```
IongIer: this { doesn't seem to have a corresponding }

king-of-the-hive:
@klautcomputing @IongIer all done, back to UTC!

PR #660, opened 2025-11-06
site stats page alpha

king-of-the-hive:
Just a very simple rough first try for the site stats page.

If you guys can try on live data and post the screenshot would be great. Not in the pages directory yet akin to puzzles and whatnot.

I have some ideas on what else can be added but I need think how to get some more fake data in my local db...

image


IongIer:
Screenshot_15
Screenshot_14

A few quick thoughts:
1. Grey text doesn't work well in darkmode.
2. The wins by color table should probably filter out casual games if it doesn't already.
3. The White > 0 and Black > 0 are confusing to me and the data looks off, what do they represent and how come they don't have the same number of games?
4.Also the numbers don't seem to sum up properly. 35 total white wins but adding em up in the table leads to less, same for the black ones. Is it because the table actually only sums up rated games whereas the totals include casual?....yeah this seems to be the case, the totals should probably mention % of which are rated or something like that.

As for what to add. Not sure would need to think about it.

db/src/models/game.rs
```diff
@@ -1209,4 +1241,140 @@ impl Game {
...
+        let cutoff_date = period_enum.cutoff_date();
+
+        let bot_filter = if !include_bots {
+            "AND black_id NOT IN (SELECT id FROM users WHERE bot = true)
+                AND white_id NOT IN (SELECT id FROM users WHERE bot = true)"
+        } else {
+            ""
+        };
+
+        let query = format!(r#"
```
IongIer: We never use raw sql, but diesel functions. That said this might be hard to achieve without raw sql but give it a try

IongIer:
I really dislike the huge type and function names like SiteStatisticsWinrateByRatingDifferenceResponse, totally a taste thing tho dunno if felix agrees.
Also this seems like a lot more code than I would expect is needed for this feature but I could be wrong.

king-of-the-hive:
I guess I could move all the aggregate functions out of Game or User objects directly into the response file. Not that it actually reduces the amount of code but at least keeps the objects free of complex functions.

I will reduce the names too.

Other than that I don't see a lot of potential for the code to be more concise. It does a lot of aggregations to display a ton of data...

apis/src/components/atoms/site_statistics_generic_card.rs
```diff
@@ -0,0 +1,124 @@
...
+
+                            view! {
+                                <tr class="border-b border-gray-200 last:border-none">
+                                    <td class="py-1">
+                                        {if first_column_is_username {
+                                            let username_for_resource = label.clone();
+                                            let label_for_fallback = label.clone();
+                                            let label_for_error = label.clone();
+                                            
+                                            let user_resource = LocalResource::new(move || {
```
IongIer: Maybe instead of looping over n users here and doing another db call for each, the response for this type should contain the complete user data (get a vector of users from the user model based on what the game model returns and that way it's two queries not n+1)

apis/src/components/atoms/site_statistics_selectors.rs
```diff
@@ -0,0 +1,167 @@
...
+    view! {
+        <div class="flex gap-2 mb-4 flex-wrap">
+            {periods
+                .into_iter()
+                .map(|period| {
+                    view! {
+                        <button
+                            type="button"
+                            class=move || {
+                                if selected_period.get() == period {
```
IongIer: In this case this is fine because SiteStatisticsTimePeriod is just a plain enum but if were anything more complex you want to use .with and that way you do the comparison without cloning the value.
```rs
selected_period.with(|p| p == period)
```

klautcomputing (changes requested):
I only skimmed this, there is some stuff I would like to change. Most importantly there should not be changes to the engine, the raw SQL queries (this is why we use an ORM) and the VeryLongJavaLikeNames.
I will give this a better look tomorrow with Ion and leave you some more detailed comments.

IongIer:
> I understand the raw SQL concern but it would be significantly more complex code and instead of pulling aggregates it would load thousand of games into memory without any substantial benefit.
> 
> The long names I shortened some already, if you have any specific you don't like I can F2 no problem :)

More complex maybe but diesel can express the queries you wrote and writing them in diesel wouldn't mean loading the games in memory, it can load exactly what the raw sql is loading.

IongIer:
One thing that is missing is this /statistics page is never exposed as a link anywhere internally. There should be buttons in the dropdowns leading to the page.

apis/src/pages/statistics.rs
```diff
@@ -0,0 +1,627 @@
...
+                                            stats
+                                                .iter()
+                                                .filter(|s| s.speed == selected_speed.get())
+                                                .cloned()
+                                                .collect::<Vec<_>>()
+                                        });
+
+                                        view! {
+                                            <h1 class="text-xl font-bold text-center">
+                                                "Player ratings distribution"
```
klautcomputing: I don't understand why it's always 1.0, this doesn't make sense to me, doesn't it need to add up to 1 in total? And 1.0 is 100%?
Image

apis/src/components/atoms/rating_buckets_chart.rs
```diff
@@ -0,0 +1,66 @@
...
+    let graph_height = Signal::derive(move || window_size.height.get() * 0.3);
+
+    view! {
+        <div class="dark:[&_text]:!fill-gray-200">
+            <Show
+                when=move || !data.get().is_empty()
+                fallback=|| {
+                    view! {
+                        <div class="text-center py-8 text-gray-500 dark:text-gray-400">
+                            "No data available"
```
klautcomputing: It's weird that the layout changes and that there's not even a heading anymore when you don't have data available.

Image

klautcomputing (changes requested):
@IongIer and I checked out the feature and we have some feedback for you on how it work. 
I am going to look at the code a bit more when it's working nicely :)

apis/src/pages/statistics.rs
klautcomputing: All users have 50+ games, also then the scale doesn't make sense because there can't be fractional users :)
I am gonna retry with more users and games and see whether I can get different results.

king-of-the-hive: It should look something like this with enough rankable users. The graph is fractional only because of super small test db.

image

engine/src/game_type.rs
```diff
@@ -16,6 +16,71 @@ pub enum GameType {
     MLP,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Copy, Default)]
```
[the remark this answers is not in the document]
king-of-the-hive: > I might not have expressed myself clearly enough, but when I said add methods to the types, I didn't also mean extend an enum with an extra variant for example, like GameSpeed:AllSpeeds. I was trying to say use what's already there from that type instead of using special strings to achieve your goal.

If the speeds should be passed as a type between the db through responses to the component as you suggested then I'm not sure how to do it without adding to the type itself. I need AllSpeeds for almost all aggregates where you see UNION ALL. I'd really appreciate specific suggestions instead of just "you can't do this" comments. 

(This also extends to the raw SQL comments: I checked the documentation and different GPTs, there's no way to have multiple group bys by calculated columns. They themselves admit that in their readme: https://github.com/diesel-rs/diesel?tab=readme-ov-file#raw-sql . If you have a specific way in mind on how to do it, please let me know, I'm happy to rewrite it. )

apis/src/components/atoms/rating_buckets_chart.rs
king-of-the-hive: yeah I guess I can try to make a common floating selector and that would actually allow to get rid of the "Games by type and speed" table, the speed selector will just update the single stat cards. And will fix the headers when data is missing.

PR #663, opened 2025-11-18
extend Swiss tournament to 64 players; add autoskip bye player 1st round

king-of-the-hive:
- Extend Swiss to 64 players since apparently we're popular
- And since we're on it, in the previous diff only added the autoskip for the Bye player to the next rounds, so first round was manual.

Would appreciate much if this can be shipped before Sunday to help Frasco out but if not he can handle

Creation:

image

Created in the Future section:

image

Skipped Bye player when odd number of players:

image


IongIer:
This looks good to me, @klautcomputing what do you think?

PR #664, opened 2025-11-18
Adds uhp string parsing to the engine, fixes parsing bug for absolute positions

IongIer:


engine/src/game_type.rs
```diff
@@ -52,3 +52,35 @@ impl FromStr for GameType {
         }
     }
 }
+
+impl GameType {
+    pub(crate) fn include_m(self) -> Self {
```
klautcomputing: Very nitpicky, but can we rename them to add_m/l/p

engine/src/history.rs
```diff
@@ -239,4 +217,226 @@ impl History {
...
+                self.game_type = self.game_type.include_l();
+            }
+            (Some('w') | Some('b'), Some('P')) => {
+                self.game_type = self.game_type.include_p();
+            }
+            _ => {}
+        }
+    }
+
+    fn parse_uhp_state(token: &str) -> Option<GameResult> {
```
klautcomputing: Wouldn't this be `parse_uhp_game_result`?

engine/src/history.rs
```diff
@@ -239,4 +217,226 @@ impl History {
...
+        playback_history.result = GameResult::Unknown;
+        match State::new_from_history(&playback_history) {
+            Ok(state) => {
+                history.result = match state.game_status {
+                    GameStatus::Finished(result) => result,
+                    _ => GameResult::Unknown,
+                };
+            }
+            Err(err) => {
+                let mut partial_history = history.clone();
```
klautcomputing: Do we need to the clone here? cant we just modify `history` itself right away?

engine/src/state.rs
```diff
@@ -155,13 +155,15 @@ impl State {
...
-                            self.play_turn(piece, target_position)?;
-                        }
-                    }
+                    let destination_piece = Piece::from_str(position)?;
+                    let target_position = self
+                        .board
+                        .position_of_piece(destination_piece)
+                        .ok_or_else(|| GameError::ParsingError {
+                            found: position.to_string(),
+                            typ: "destination piece on board".to_string(),
```
klautcomputing: "not on board"? Or am I getting something mixed up here?

klautcomputing (changes requested):
This was fun to look at, I might have left you too many nitpicky comments :)

engine/src/history.rs
IongIer: NotStarted|InProgress are not really  results. maybe game_status?

PR #678, opened 2026-01-13
update leptos_i18n to v0.6

Baptistemontan:


klautcomputing:
Thanks for this! There's a lot of other dependencies that need updating as well, so I made a feature branch and will merge your PR into this and then once everything works on leptos 0.8.15 we merge it into main.

Thanks again for your work 🙂

PR #694, opened 2026-01-31
Fixes challenge visual overlap

IongIer:
Is this what we want or some other way to fix?
A fix for #709 along with some nicer code


KES777:
@IongIer Sorry, can not review =(
Though if you post screenshot, I could.

Is there a domain for testing? Probably I could try it.

PR #699, opened 2026-02-07
Hivespider's video

klautcomputing:
Take a look whether you like it, it definitely needs to be shrunk down to 1/4 of its size.


klautcomputing:
Why was this autoclosed?

apis/src/components/organisms/featured_video.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    // auth.user is async (derived from a server action), so is_logged_in()
+    // starts as false and resolves later. Gating `show` on it causes a flash:
+    // dismissed users briefly appear logged-out, which forces show=true.
+    // Cookie state is synchronous so we only depend on that for visibility.
+    let show = move || !config().video_dismissed;
+
+    view! {
+        <Show when=show>
+            <div class="w-full m-2 flex flex-col items-center">
```
IongIer: "flex overflow-hidden justify-center m-2 w-full lg:justify-end 2xl:justify-center"

apis/src/components/organisms/featured_video.rs
```diff
@@ -0,0 +1,72 @@
...
+                <div class="w-full max-w-screen-md overflow-hidden rounded-lg border border-gray-200 dark:border-gray-700 bg-white dark:bg-gray-800 hover:border-gray-300 dark:hover:border-gray-600 transition-colors">
+                    <div class="flex items-center justify-between px-3 pt-3 mb-3 sm:px-4 sm:pt-4">
+                        <h2 class="text-lg font-bold text-gray-800 dark:text-gray-200">
+                            {t!(i18n, home.learn_to_play)}
+                        </h2>
+                        {move || is_logged_in().then(|| view! {
```
IongIer: We should probably allow logged out users to dismiss it

apis/src/components/organisms/featured_video.rs
```diff
@@ -0,0 +1,72 @@
+use crate::{i18n::*, providers::{AuthContext, Config}};
+use leptos::prelude::*;
+
+const VIDEO_URL: &str = "https://www.youtube.com/watch?v=-_CT8cgOR5Q";
+const THUMBNAIL_URL: &str = "https://img.youtube.com/vi/-_CT8cgOR5Q/hqdefault.jpg";
```
IongIer: Maybe we want this as an asset so we don't request this whenever someone loads the page and leak that info to yt as well as it shows up as referrer hivegame.com

apis/src/pages/home.rs
```diff
@@ -37,6 +40,7 @@ pub fn Home() -> impl IntoView {
                     </div>
                 </div>
                 <div class="flex flex-col order-2 items-center space-y-6 min-w-0 lg:order-none lg:col-start-2 lg:row-start-1">
+                    <FeaturedVideo />
```
IongIer: This being above quickplay means that someone opening the site on mobile with an admin banner active will see just the banner and the video thumbnail (and even that might be partly below the fold)...although that might be fine. just something to consider

PR #703, opened 2026-03-13
use a Store for the GameState context

PenguinWithATie:
- Optimizes the game signal into a store so fields are updated only when they themselves are modified.

#### Requires https://github.com/hiveboardgame/hive/pull/704


apis/src/providers/game_state.rs
```diff
@@ -194,47 +199,51 @@ impl GameStateSignal {
...
+            })
         })
     }
 
     pub fn is_first_turn_as_signal(&self) -> Signal<bool> {
-        create_read_slice(self.signal, |gs| {
-            gs.history_turn.is_none() || gs.history_turn == Some(0)
+        let game_state = self.0;
+        Signal::derive(move || {
+            game_state.with(|gs| gs.history_turn.is_none() || gs.history_turn == Some(0))
```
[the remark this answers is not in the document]
PenguinWithATie: true, I have tried to fix most with calls to call the inner fields properly now

apis/src/components/organisms/chat.rs
```diff
@@ -88,12 +92,12 @@ pub fn ChatWindow(
...
     let chat = expect_context::<Chat>();
     let auth_context = expect_context::<AuthContext>();
-    let game_state = expect_context::<GameStateSignal>();
+    let game_state = expect_context::<GameStateStore>();
     let uid = auth_context
         .user
         .with_untracked(|a| a.as_ref().map(|user| user.user.uid));
-    let white_id = create_read_slice(game_state.signal, |gs| gs.white_id);
-    let black_id = create_read_slice(game_state.signal, |gs| gs.black_id);
+    let white_id = Signal::derive(move || game_state.white_id().get());
```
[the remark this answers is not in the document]
PenguinWithATie: I agree its better to have just a simple closure. But in general, what is the logic/stylistic guide to prefer a plain closure over a derive? From leptos POV they seem to be more or less the same thing

apis/src/components/organisms/chat.rs
```diff
@@ -88,12 +94,12 @@ pub fn ChatWindow(
...
     let chat = expect_context::<Chat>();
     let auth_context = expect_context::<AuthContext>();
-    let game_state = expect_context::<GameStateSignal>();
+    let game_state = expect_context::<GameStateStore>();
     let uid = auth_context
         .user
         .with_untracked(|a| a.as_ref().map(|user| user.user.uid));
-    let white_id = create_read_slice(game_state.signal, |gs| gs.white_id);
-    let black_id = create_read_slice(game_state.signal, |gs| gs.black_id);
+    let white_id = move || game_state.white_id().get();
```
IongIer: You do not need the move here is what I was saying. Exactly this should work:
```rs
let white_id = game_state.white_id();
let black_id = game_state.black_id();
let white_id = move || white_id.get().expect("Game has white player");
let black_id = move || black_id.get().expect("Game has black player");```

No closure for the first set of white_id = and then a .get() instead of a plain () before the expect

File: apis/src/components/molecules/control_buttons.rs [hunk 4]
```diff
@@ -32,25 +32,26 @@ pub fn ControlButtons() -> impl IntoView {
...
     });
-    let pending = create_read_slice(game_state.signal, |gs| gs.game_control_pending.clone());
-    let not_tournament = create_read_slice(game_state.signal, |gs| {
-        gs.game_response
-            .as_ref()
+    let pending = Signal::derive(move || game_state.game_control_pending().get());
+    let not_tournament = Signal::derive(move || {
+        game_state
+            .game_response()
+            .get()
```
IongIer: with()

apis/src/components/molecules/rating_and_change.rs
```diff
@@ -32,9 +35,11 @@ pub fn RatingAndChangeDynamic(
...
     side: Color,
 ) -> impl IntoView {
-    let game_state = expect_context::<GameStateSignal>();
-    let ratings = create_read_slice(game_state.signal, |gs| {
-        gs.game_response
+    let game_state = expect_context::<GameStateStore>();
+    let ratings = Signal::derive(move || {
+        game_state
+            .game_response()
+            .get()
```
IongIer: with()

apis/src/providers/game_state.rs
```diff
@@ -16,59 +17,67 @@ use super::{
...
+        let color_allowed = |color: &Color, game_response: &GameResponse| {
+            let rated = game_response.rated;
+            let takeback = match color {
+                Color::Black => &game_response.black_player.takeback,
+                Color::White => &game_response.white_player.takeback,
+            };
+            takeback == &Takeback::Always || takeback == &Takeback::CasualOnly && !rated
+        };
+        self.game_response()
+            .get()
```
IongIer: with()

PR #705, opened 2026-03-14
Analysis improve

### LINKED ISSUE #508: After 30 min of analysis the page becomes too slow on mobile.

### LINKED ISSUE #691: Don't override browser shortcuts in analysis mode
A common browser shortcut to go back in history is `alt+leftArrow`. Being able to use this could be handy to go back to a live game after jumping in analysis mode.  
For now, it just goes back in the history of the game (like simply `leftArrow` would), shadowing this keyboard shortcut (for no good reason IMO).

### LINKED ISSUE #696: Analysis slow and move indicator not visible
Hello! Two issues in the Analysis tool:

- when the game to analyse is quite long and you have been adding alternative variations for a while, it becomes pretty slow, almost freezes sometimes. I assume it's just due to the amount of data, cause it never happens for short games or at the beginning of an analysis as far as I remember. This for desktop (Chrome), not sure about mobile.

- When analysing, the indicator of the move in the history (the coloured rectangle that highlights the move) disappears. That is, the notation of the move you are looking at gets pushed to the bottom and it's not visible anymore.
The rectangle should always be visible, at the top somewhere in the middle I guess. 
Right now, when you scroll the notation to make it visible, as soon as you click on the arrow to go to the next move, the rectangle goes outside of the visible area. Interestingly, when you click the arrow to go to the previous move, the rectangle gets visible again! (the second-last line seems to me). But when you click the arrow to go to the beginning of the game, the problem is there too. 
It's a small detail but I think it would make the experience smoother!

Image

PenguinWithATie:
- Uses a Store for analysis

- ~Loads Analysis with a OnceResource and avoids the suspend barrier~ Use a local resource, no reason no load this in the server. Do a simple refactor for readability but the suspend barrier when providing the context makes sense and should be required (ion is always right :D)

- Ensures uhp strings and the move param in the query can be used in tandem

- Only move in analysis history if arrow is pressed without any modifiers. (Fixes https://github.com/hiveboardgame/hive/issues/691)

### Created the full_path field in Analysis tree
This maintains a full path from root to leaf of a "best guess" for the possible variation the user wants to traverse. Then instead of redrawing everything, jump forward or backward a few moves. It has the following advantages:
- The arrows work fast since a single step is very quick to do.
- Jumping to "near" points in the tree is quick
- In the worst case we need to recompute the full path, which wont happen if the tree isnt to dense.  And even then, recomputing the full path for a 150~200 nodes is not that bad.
- Its not required to serialized and can be reconstructed from the rest of the data

#### Requires https://github.com/hiveboardgame/hive/pull/703

Hopefully fixes https://github.com/hiveboardgame/hive/issues/508 and https://github.com/hiveboardgame/hive/issues/696

### DIFF
--- a/apis/src/components/atoms/active.rs
+++ b/apis/src/components/atoms/active.rs
@@ -1,6 +1,6 @@
 use crate::{
     common::{ActiveState, SvgPos},
-    providers::game_state::GameStateSignal,
+    providers::game_state::GameStateStore,
 };
 use hive_lib::Position;
 use leptos::{either::Either, prelude::*, text_prop::TextProp};
@@ -13,7 +13,7 @@ pub fn Active(
     active_state: ActiveState,
     straight: bool,
 ) -> impl IntoView {
-    let mut game_signal = expect_context::<GameStateSignal>();
+    let game_signal = expect_context::<GameStateStore>();
     let center = move || SvgPos::center_for_level(position, level, straight);
     let transform = TextProp::from(move || format!("translate({},{})", center().0, center().1));
     match active_state {
--- a/apis/src/components/atoms/download_pgn.rs
+++ b/apis/src/components/atoms/download_pgn.rs
@@ -4,22 +4,25 @@ use leptos_icons::*;
 use wasm_bindgen::{JsCast, JsValue};
 use web_sys::{js_sys::Array, Blob, Url};
 
-use crate::{providers::game_state::GameStateSignal, responses::GameResponse};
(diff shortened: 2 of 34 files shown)


IongIer:
decided to get rid of tree-ds completly instead of rebasing this on main. #784

PR #712, opened 2026-03-31
Fix pwa style and caching

IongIer:
Refreshing on a game url on a mobile browser fits everything correctly, doing the same thing on a installed mobile PWA pushes the bottom reserve below the viewport leading to needing to scroll down. 

This fixes that issue alongside making the pwa caching less manual.


PenguinWithATie (commented):
Cant comment much on the JS part as I dont have much experience.

klautcomputing (approved):

PR #719, opened 2026-04-27
Adds admin button to stop real time game creation

klautcomputing:


IongIer:
Realtime tournaments can still start via jobs::tournament_start/Tournament::automatic_start or
TournamentAction::Start, which call Game::create(NewGame::new_from_tournament(...)) without reading
realtime_games_enabled. Gate those tournament start paths on the same setting.

PR #726, opened 2026-04-29
Smarter pass detection

IongIer:


engine/src/board.rs
```diff
@@ -622,6 +685,10 @@ impl Board {
         Piece::new_from(Bug::from(bug), Color::from(color), order)
     }
 
+    fn slot_represents_piece(&self, offset: usize, game_type: GameType) -> bool {
```
klautcomputing: nitpick: slot -> offset

engine/src/bug.rs
```diff
@@ -321,6 +331,12 @@ impl Bug {
         }
     }
 
+    fn has_ant_move(position: Position, board: &Board) -> bool {
```
klautcomputing, line 338: Shouldn't this just be 
```rs
Bug::crawl(position, board).next().is_some()
```
 because if the ant can't move one space, it cannot move anywhere.

klautcomputing (approved):
two small things

PR #748, opened 2026-05-15
Add keyboard arrow-key history navigation to the game page

danielbairddev:
- ArrowLeft from anywhere on the game page immediately enters history mode and steps back one move (lichess-style). Subsequent presses continue stepping back.
- ArrowRight steps forward when already browsing history.
- Removes the old per-component keyboard listener in history_controls, which only fired when the History tab was already open.
- Uses leptos_use SSR-safe wrappers (use_window/UseDocument) so the event listener is a no-op during server-side rendering.
- When the user is viewing the latest history move and an opponent move arrives, automatically snaps back to live view so they can respond without manually navigating forward.

https://github.com/user-attachments/assets/4e8c3e1c-49f1-4e35-b1e4-9719586bf692

https://github.com/user-attachments/assets/79b8bd10-0ebb-4b6b-b1fb-4d4263d41dce

### DIFF
--- a/apis/src/components/atoms/history_button.rs
+++ b/apis/src/components/atoms/history_button.rs
@@ -1,8 +1,13 @@
 use crate::{
     components::organisms::side_board::move_query_signal,
-    providers::game_state::GameStateSignal,
+    providers::game_state::{GameStateSignal, View},
 };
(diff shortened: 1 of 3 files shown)


klautcomputing (approved):
Nice I like it 🙂

PR #750, opened 2026-05-27
Adds hashes to games and creates game_hashes table

klautcomputing:
This PR:
1. adds a script to backfill hashes on the games table via cd db; cargo run --bin backfill_hashes;
2. a new table: game_hashes which is a lookup table which can be used to answer questions like: show me the top X games where this hash appeared (and more complex queries in the future)


klautcomputing:
Will get pulled into main in #760

PR #753, opened 2026-05-31
Smoother zoom

klautcomputing:
Makes zoom a lot smoother, pinch works as well


klautcomputing:
Before we merge this #749 needs to be merged!

PR #764, opened 2026-06-11
Adds migration fix script to flake

klautcomputing:
@IongIer please take a look at this

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,6 +31,219 @@
           rev = "master";
           sha256 = "sha256-7Q2VuVexug0iqBXEzHfQ/c9q7TfjL56psGbq5sU2Nw4=";
         }) { };
+        # Migration drift tool. Two entry points:
+        #   on-checkout PREV NEW FLAG  - post-checkout hook; reverts migrations
+        #                                the new branch lacks but the DB applied.
+        #   fix [--dry-run]            - manual repair; reverts every migration
+        #                                in the DB that the current branch lacks.
+        # A migration's down.sql is not on disk once you've left its branch, so we
+        # recover it from git (the prior commit, or any branch/tag for `fix`),
+        # run it, then drop the row from diesel's ledger.
+        migrations-tool = pkgs.writers.writePython3Bin "migrations-tool" { flakeIgnore = [ "E501" ]; } ''
+          import argparse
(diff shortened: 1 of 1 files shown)


IongIer:
TLDR:   
  1. Stop after any failed or unavailable rollback, including missing down.sql.
  2. Preserve an existing post-checkout hook.
  3. Compare migrations by Diesel’s numeric version, not directory name.
  4. Reject ambiguous down.sql candidates instead of choosing the first ref.
  5. Avoid installing a potentially GC-broken /nix/store symlink as the persistent hook.

As i don't know nix at all....BEEP BOOP I AM A ROBOT:
  - [P2] Stop reverting after a down migration fails — flake.nix:154-156
    When reverting multiple orphaned migrations, a failure here only emits a warning and the caller proceeds to
    older migrations. If a newer migration cannot be reverted but an older dependency can, the database is left
    with the newer migration applied while its prerequisite and ledger entry are removed. Propagate the failure
    and stop processing older migrations.

  - [P2] Preserve existing post-checkout hooks — flake.nix:313-314
    When a developer already has a post-checkout hook, ln -sf replaces it without warning, silently disabling any
    existing tooling attached to that hook. The installer should refuse installation, preserve and chain the
    existing hook, or use a hook-management mechanism rather than overwriting it.
      
   - [P2] Compare migrations by Diesel version, not directory name — flake.nix:184
    only_old compares full directory names, but Diesel identifies migrations solely by their numeric version. If
    a migration is renamed while retaining its timestamp/version—or two branches use different names for the same
    version—the hook treats the old directory as removed, executes its down.sql, and deletes the ledger row even
    though the new branch still contains that version.

  - [P2] Do not choose an arbitrary migration when versions collide — flake.nix:166
    locate() accepts the first branch or tag containing a requested version. Independent branches can contain
    different migrations with the same timestamp/version, so migration-fix may execute an unrelated down.sql. It
    should detect multiple distinct candidates and refuse or require an explicit choice.

  There is also a lower-severity durability concern: flake.nix:314 is a symlink into /nix/store. It can become
  broken after garbage collection or after the dev-shell generation changes because repository symlinks are not
  Nix GC roots. This could be addressed alongside the existing “preserve existing hook” finding.

PR #784, opened 2026-07-24
Improves analysis performance

IongIer:
This is a rewrite of analysis away from tree-ds.


klautcomputing (approved):
Engine changes and click through look fine to me

PR #820, opened 2026-09-06
Fixes Position out of bounds

klautcomputing:
We need to make sure serde deserialize also goes through Position::new so we stay inside the torus. This is a non issue for the website as both sides are our engine, but as soon as we let bot devs send via WS they could have triggered this.


klautcomputing:
Superseded by off-world, which already guards this at the State boundary and rejects an off-board target rather than normalizing it.

PR #821, opened 2026-09-07
Fix opening explorer notation to match current board orientation

Moltenfire:
The explorer showed a suggested move notation exactly as the server stored it, even though that notation was recorded under a different game's board orientation. Notation is now regenerated locally from the simulated move against the viewer's own board instead.

Tested on local instance.


klautcomputing:
Thank you! <3

klautcomputing (approved):

